PR #21, opened 2015-08-29
Fix readme doc link

andrewgrz:
Fix the link to point to the correct address. It was pointing relatively before.


almarklein:
Thanks!

PR #30, opened 2015-09-05
Fixed paths to Chrome and Chromium browsers on Linux

### LINKED ISSUE #29: Flexx looking for Chrome on linux in /usr/lib
When trying to use the Chrome runtime on Linux I was getting:

    RuntimeError('Chrome or Chromium browser was not detected.')

chromeapp.py is looking for Chrome and Chromium in /usr/lib for some reason:

```
elif sys.platform.startswith('linux'):
    paths.append('/usr/lib/google-chrome/google-chrome')
```

I don't know of any distros that will be keeping the browser binaries in /usr/lib. Furthermore, the Chrome binary filename should also include the channel. That is: google-chrome-stable, google-chrome-beta, google-chrome-dev.

I fixed the path in chromeapp.py and it's working great.

Changing the above code to:

```
elif sys.platform.startswith('linux'):
    paths.append('/usr/bin/google-chrome-stable')
    paths.append('/usr/bin/google-chrome-beta')
    paths.append('/usr/bin/google-chrome-dev')
```

...should support the vast majority of modern distros and all of the Chrome release channels one may have installed.

The same fix should be applied to the Chromium detector function with appropriate file names.

HittingSmoke:
I posted this on the issue tracker but I realized despite never having made a pull request before, I knew enough about git to figure it out so I made one.

The links to Chrome and Chromium on Linux are currently pointing to /usr/lib/google-chrome/google-chrome and /usr/lib/chromium-browser/chromium-browser.

The correct paths on modern distros to Chrome and Chromium are /usr/bin/google-chrome-$CHANNEL and /usr/bin/chromium respectively.

Tested on Arch Linux with Google Chrome stable channel.

closes #29

### DIFF
--- a/flexx/webruntime/chromeapp.py
+++ b/flexx/webruntime/chromeapp.py
@@ -25,7 +25,9 @@ def get_chrome_exe():
         paths.append(os.path.expanduser("~\\AppData\\Local\\Google\\Chrome\\chrome.exe"))
         paths.append(os.path.expanduser("~\\Local Settings\\Application Data\\Google\\Chrome\\chrome.exe"))  # xp
     elif sys.platform.startswith('linux'):
-        paths.append('/usr/lib/google-chrome/google-chrome')
+        paths.append('/usr/bin/google-chrome-stable')
+        paths.append('/usr/bin/google-chrome-beta')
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks!

PR #36, opened 2015-09-20
Removed a few instances of asserts.

kevinlondon:
I changed a few of the assertions to use explicit errors in case of invalid input / checks.

I was using [Bandit](https://github.com/openstack/bandit) to check for potential vulnerabilities in a few popular Python projects and this came up as a warning:

```
>> Issue: Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
   Severity: Low   Confidence: High
   Location: flexx/webruntime/common.py:23
19	    """ Base class for all runtimes.
20	    """
21
22	    def __init__(self, **kwargs):
23	        assert 'url' in kwargs
24	        self._kwargs = kwargs
25	        self._proc = None
26	        self._streamreader = None

>> Issue: Use of assert detected. The enclosed code will be removed when compiling to optimised byte code.
   Severity: Low   Confidence: High
   Location: flexx/webruntime/xul.py:495
491	        # Get app of firefox
492	        if not 'Contents/MacOS' in xul_path:
493	            raise NotImplementedError('Cannot deal with real xulrunner yet')
494	        xul_app = op.dirname(op.dirname(op.dirname(xul_path)))
495	        assert xul_app.endswith('.app')
496
497	        # Clear destination
497	        # Clear destination
498	        if op.isdir(path):
498	        if op.isdir(path):
```

As per the warning, `assert` statements are stripped out if compiling with optimized bytecode (-O option) in Python. Here's a [StackOverflow question](http://stackoverflow.com/questions/2830358/what-are-the-implications-of-running-python-with-the-optimize-flag) with a little discussion and the original [Python docs](https://docs.python.org/2/reference/simple_stmts.html#the-assert-statement).

In these cases, they're mostly forward-looking sanity checks on the input. Still, it's best to replace asserts with other things, especially if there's a possibility that it would be compiled with the -O option. I found a number of other instances where asserts were used but these felt like the most prudent. I'll try to come back and swap the rest at a later time.

This commit also strips trailing whitespace in a few files. Apologies for the noise that adds to the PR.

### DIFF
--- a/flexx/webruntime/common.py
+++ b/flexx/webruntime/common.py
@@ -23,7 +23,6 @@ class BaseRuntime(object):
         if not 'url' in kwargs:
             raise KeyError('No url provided for runtime.')
 
-        assert 'url' in kwargs
         self._kwargs = kwargs
         self._proc = None
         self._streamreader = None
--- a/flexx/webruntime/common.py
+++ b/flexx/webruntime/common.py
@@ -18,17 +18,20 @@ from ..util.icon import Icon
 class BaseRuntime(object):
     """ Base class for all runtimes.
     """
-    
+
     def __init__(self, **kwargs):
+        if not 'url' in kwargs:
+            raise KeyError('No url provided for runtime.')
+
         assert 'url' in kwargs
         self._kwargs = kwargs
         self._proc = None
         self._streamreader = None
         atexit.register(self.close)
-        
+
         logging.info('launching %s' % self.__class__.__name__)
         self._launch()
-    
+
     def close(self):
         """ Close the runtime, or kill it if the process does not
         respond. Note that closing does not work when the runtime is a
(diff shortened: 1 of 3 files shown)


flexx/webruntime/common.py
```diff
@@ -18,17 +18,20 @@
 class BaseRuntime(object):
     """ Base class for all runtimes.
     """
-    
+
     def __init__(self, **kwargs):
+        if not 'url' in kwargs:
+            raise KeyError('No url provided for runtime.')
+
         assert 'url' in kwargs
```
almarklein: Can you remove the now redundant assert?

almarklein:
Thanks. You make a good point. I think its just me being a bit lazy during initial development. I plan to do a good review of all the source sometime soon, and this is one of the things that I'd need to fix.

kevinlondon:
Yep! Added, thank you for the catch. By the way, I don't think this reflects poorly on you at all. Everyone misses things. Before I started using Bandit in preparation for a talk, I didn't know about the assert stripping.

almarklein:
Thank you!

PR #37, opened 2015-09-30
Use Phosphor layouts

almarklein:
This makes flexx.ui be based on Phosphor, so we can take advantage of the great layout and future widgets. Also fixes several other things on the way.


mgaitan:
Hurray!

PR #41, opened 2015-10-16
Assets

almarklein:
* Change how assets are handled. Making things more efficient. Making a difference between global assets and session-specific assets. Also allows adding assets e.g. for widgets that need this.
* This lead to a refactoring of the whole app subpackage
* Everything related to Tornado is now in one module
* Add example with jQuery
* Rename Proxy class to Session
* Rename Pair class to Model


datnamer:
Cool :)

PR #47, opened 2015-10-19
test travis size check

almarklein:


almarklein:
Hooray, Travis' size check works!

PR #51, opened 2015-10-30
Commonast

almarklein:
This implements `commonast.py`, which can produce "normalized" ast, so that the pyscript transpiler can be a bit simpler and does not have to care about support for different Python versions.

If we ever implement an pure Python ast parser, we can target the same ast format.

Note that backward compatibility can be tough: a new Python may introduce syntax that we cannot handle with this ast format. In such an event, we need to revise the standard and update PyScript to make use of new format.


almarklein:
the only travis error is the size check, which is expected with this size pr

PR #70, opened 2015-12-07
Tweaks to avoid leaks

### LINKED ISSUE #62: Session memory leak
There is a memory leak, because the demo server http://52.21.93.28:8000/ keeps accumulating RAM usage as it has handled  more sessions.

almarklein:
* `session.status` is a bit easier, you can now check for it to evaluate to Falsy to determine if the connection is closed.
* Fixed leak in chatroom example
* Old pending sessions are cleared once in a while

I hope this fixes #62

### DIFF
--- a/examples/ui/chatroom.py
+++ b/examples/ui/chatroom.py
@@ -46,6 +46,8 @@ class ChatRoom(ui.Widget):
         self._update_participants()
     
     def _update_participants(self):
+        if not self.session.status:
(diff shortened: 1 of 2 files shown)


almarklein:
AWS server is running this branch. If memory is constant after a few hours today I consider it fixed. (and then we'll leave it running to see what the long term looks like).

almarklein:
so far so good

PR #79, opened 2016-01-18
Support for Python 2.7

### LINKED ISSUE #33: Support legacy Python (py 2.7)?
Via six.py or python-future.org

I'm a bit tired of making stuff work on both Python2 and Python 3. I just want to write `isinstance(c, str)`. But maybe to get Flexx out there and usable from existing projects, I'd have to give in. We'll see.

PyScript support is the trickiest bit, I suppose, because the ast parser behaves slightly different. But the ast parse of pypy is different too, and I very much like to support that. Having a pure Python ast parser would be even more awesome, see #19.

### LINKED ISSUE #50: In flexx.react can we use function annotations to define types or set default values?
Now:

    @react.input
    def foo(value=2):
        return float(value)

Maybe:

    @react.input
    def foo(value:2):
        return float(value)

... which would enable getting the default value without executing the function.

or:

    @react.input
    def foo(value:float=2):
        return float(value)

almarklein:
Closes #33 and closes #50

Going for an approach where we translate the codebase to Python 2.x during the installation step. Much like `2to3`, but then `3to2`.


almarklein:
Running `lib3to2` on the codebase, I see the following changes that applied to our code:

* Inclusion of  `from __future__ import absolute_import, print_function, with_statement`
* Put an `u` in front of multiline strings
* Put an `u` in front of regular strings
* Fix imports of `urlopen`  and related functions.
* `str` -> `unicode` and `chr` -> `unichr`
* `class X:` -> `class X(object):`
* `super()` -> `super(Cls, self)`
* `range()` -> `xrange()`
* `__self__` -> `im_self` (but we should probably not use `__self__` in the way we do now)

And some that we don't need if we stick to Python 2.7:
* `b''` -> `''`
* `bytes` -> `str`
* `from io import open()`
* use of `izip` and `ifilter`

pya:
Did have a look at http://python-future.org ? It allows to run Python 3 code in Python 2 without a translation step. Adding a bunch of imports at the beginning of a module is enough. This would be a single-source approach. There are only very things that might not work (`nonlocal`, and `*` unpacking, I think). But it even works for `super()` and metaclasses.

almarklein:
Hi @pya, yeah I looked at that, and also python-modernize. But I hope I can make it work with a translation step at install time, without having to change the source. I have been writing single-source code that runs on Python 2 and Pytjon 3 for years, and was hoping I could just write Python 3 this time :)

PR #103, opened 2016-02-24
Update releasenotes.rst (dates)

pierrejean-coudert:


almarklein:
Habits die hard :/  Thanks for the fix!

PR #113, opened 2016-05-11
Tweaks related to new event system

almarklein:
Continuing #108

* improve mouse events
* make that in `JS.init()` you have access to sub models created in `Py.init()`
* session is implicitly inherited by models when used as a context manager. 
* There is a func to get the currently active (as in being "in context") model, that can be used in subclasses
* props can deal with numpy arrays

### DIFF
--- a/docs/app/api.rst
+++ b/docs/app/api.rst
@@ -29,6 +29,8 @@ Functions related to us Model classes as apps
 
 .. autofunction:: flexx.app.get_model_classes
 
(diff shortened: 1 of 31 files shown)


Korijn:
The changes on mouse events seem really useful!

What do the session changes offer?

almarklein:
The session changes mean that you can use models as a context manager (or define other models in the `init()`) to make them share the session, instead of having to do `MyModel(session=otherModel.session)`.

PR #119, opened 2016-05-27
Use the xdg standard to store the config in linux

Konubinix:


almarklein:
Thanks for this. I get the `XDG_CONFIG_HOME`, but have some doubts about the `os.path.expanduser(os.path.join("~", ".config"))`. Common locations are also `~/.myapp.cfg` or `~/.myapp/config.cfg`, and they would be impossible to reach by means of the `~appdata` prefix in this way. 

Is `~/.config` part of a standard? If it is, why is any home dir always so cluttered with hidden dirs?

Konubinix:
Dear Almar Klein,

Thank you for taking the time to consider this PR.

Almar Klein <notifications@github.com> writes:

> Thanks for this. I get the `XDG_CONFIG_HOME`, but have some doubts
> about the `os.path.expanduser(os.path.join("~", ".config"))`.

According to the xdg specification
(https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html),
the default value of the XDG_CONFIG_HOME variable is '~/.config/'.

> Common locations are also `~/.myapp.cfg` or `~/.myapp/config.cfg`

Indeed, I have seen these locations before, but I thought that it was
some emerging behavior due to the lack of standard. I have also seen
several projects moving the configuration files from directories like
`~/.myapp.cfg` or `~/.myapp/config.cfg` to ones following the xdg
standard.

As a user, I tend to prefer the xdg specification over the '~/.myapp/'
directory because it avoids cluttering my home directory. In addition, I
can keep all my configuration files in a predefined place by setting
only one environment variable: XDG_CONFIG_HOME.

> , and they would be impossible to reach by means of the `~appdata`
> prefix in this way.

IMHO, user using the ~appdata prefix want the framework to use a
standard directory, not the '~/.myapp/' folder.
(like get_app_dir in click ?
http://click.pocoo.org/5/api/#click.get_app_dir).

My best,
-- 
Konubinix
GPG Key    : 7439106A
Fingerprint: 5993 BE7A DA65 E2D9 06CE  5C36 75D2 3CED 7439 106A

almarklein:
Ok, I see that click does use the `.config` as a default. I'll assume that this is the best approach and that `~.config` always exists (I checked that it does on Ubuntu Lucid Lynx).

almarklein:
Thanks!

PR #121, opened 2016-05-31
Fallback on e.button if e.buttons is not available

Konubinix:
I am running uzbl, built with libjavascriptcoregtk-1.0-dev (2.4.11-1) . In this setup, the event attribute buttons is undefined whereas the attribute button exists. They both appear to have the same purpose in the context of mouse clicking.

### DIFF
--- a/flexx/ui/_widget.py
+++ b/flexx/ui/_widget.py
@@ -1,8 +1,8 @@
 """
 .. UIExample:: 100
-    
+
     from flexx import app, ui
-    
+
     class Example(ui.Widget):
         ''' A red widget '''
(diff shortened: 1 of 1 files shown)


almarklein:
I don't even know what uzbl is, but thanks for the fix! Can you please add a short comment that why the fallback is necessary, so it wont get removed later?

Konubinix:
Uzbl is a minimalist web browser built around webkit and libsoup (http://www.uzbl.org/)

I updated the pull request. Thank you for the answer.

almarklein:
Awesome, thanks!

PR #128, opened 2016-06-05
Fix locating firefox on OS X

kamoser:
Some mac installations don't link firefox into '/Applications', (e.g. homebrew links into '~/Applications'). A more reliable method is to use the app id for getting the path, (i.e. 'mdfind kMDItemCFBundleIdentifier==org.mozilla.firefox').
For path fallbacks, '~/Applications/...' added in addition to '/Applications/...'.

### DIFF
--- a/flexx/webruntime/xul.py
+++ b/flexx/webruntime/xul.py
@@ -170,7 +170,19 @@ def get_firefox_exe():
         paths.append('/usr/lib/iceweasel/iceweasel')
         paths.append('/usr/lib64/iceweasel/iceweasel')
     elif sys.platform.startswith('darwin'):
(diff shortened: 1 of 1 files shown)


almarklein:
Thanks! A few minor comments:

If you add a space after the comma on line 178, the linter will be happy :)

Instead of `environ['HOME']` can you please use `os.path.expanduser('~'), because AFAIU it does the same but is a bit more generic.

I feel a bit uncomfortable with calling a subprocess while firefox is in a "standard" place 95% of the time. Do you think it makes sense to put that code in the `else` that iterates over the possible paths? In other words: check the two standard places, and if that does not work, try using `mdfind`.

almarklein:
Any progress on this?

almarklein:
Superseded by #147

PR #147, opened 2016-07-01
Fix locating firefox on OS X (part 2)

almarklein:
Superseded #128


almarklein:
@kamoser I took your commit and modified it a bit to only do the subprocess call if necessary. Let me know if this does not work. Thanks for the fix!

PR #153, opened 2016-07-05
Refactor tornado server

### LINKED ISSUE #152: Allow more flexibility in creating apps/io loop
Requirements:

* Be able to run in a separate thread (this was enabled earlier, but I think I missed a part)
* Be able to run in multiprocessing on Windows (if possible)
* Be able to start and stop Flexx multiple times, using the same ioloop
* Dito, using differen ioloop instances
* Multiple ioloops in parallel?

cc @Maxyme

almarklein:
Fixes #152 

cc @korijn @Maxyme

### DIFF
--- a/flexx/app/examples/flexx_in_thread.py
+++ b/flexx/app/examples/flexx_in_thread.py
@@ -1,0 +1,51 @@
+"""
+Example showing running Flexx' event loop in another thread.
+This is not a recommended use in general.
+
+Most parts of Flexx are not thread-save. E.g. setting properties
+should generally only be done from a single thread. Event handlers
+are *always* called from the same thread that runs the event loop
+(unless manually called).
+
+The app.create_server() is used to (re)create the server object, binding
+it to the ioloop of the current thread. Note that app.start() must be
+called from that same thread.
+
+"""
+
+import time
+import threading
+
+from flexx import app, event
+
+
+class MyModel1(event.HasEvents):
+    @event.prop
+    def foo(self, v=0):
+        return v
+    
+    @event.connect('foo')
+    def on_foo(self, *events):
+        for ev in events:
+            print('foo changed to', ev.new_value)
+
+# Create model in main thread
+model = MyModel1()
+
+def main():
+    app.create_server()
+    model.foo = 2  # Probably a bad idea to set props from both threads
+    app.start()
+
+# Start Flexx server in new thread
+t = threading.Thread(target=main)
+t.start()
+
+# Manipulate model from main thread (the model's on_foo() gets called from other thread)
+for i in range(5, 9):
+    time.sleep(1)
+    model.foo = i
+
+# Stop event loop (this is thread-safe) and wait for thread to end
+app.stop()
+t.join()
--- a/flexx/app/funcs.py
+++ b/flexx/app/funcs.py
@@ -25,10 +25,13 @@ _current_server = None
 
 def create_server(host=None, port=None, backend='tornado'):
     """
-    Create a new server object. Flexx uses a notion of a single current
-    server object. This function creates that object. If there already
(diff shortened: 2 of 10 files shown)


flexx/app/funcs.py
```diff
@@ -11,113 +11,129 @@
...
+def get_current_server():
+    """
+    Get the current server object. Creates a server if there is none.
+    Currently, this is always a TornadoServer object, which has properties:
+    
+    * serving: a tuple ``(hostname, port)`` specifying the location being served,
+      or ``None`` if not servering yet/anymore.
+    * app: the ``tornado.web.Application`` instance
+    * loop: the ``tornado.ioloop.IOLoop`` instance
+    * server: the ``tornado.httpserver.HttpServer`` instance
```
almarklein: So `app.init().native` -> `app.get_current_server().app`

almarklein:
Now that we're in here. Maybe we can think of a better way for "server". Maybe application is a bit more accurate. Thoughts/ideas?

flexx/app/examples/flexx_in_thread.py
```diff
@@ -0,0 +1,51 @@
...
+# Create model in main thread
+model = MyModel1()
+
+def main():
+    app.create_server()
+    model.foo = 2  # Probably a bad idea to set props from both threads
+    app.start()
+
+# Start Flexx server in new thread
+t = threading.Thread(target=main)
```
Korijn: You could just say `t = threading.Thread(target=app.start)` by the looks of it, right?

Maxyme:
in the tests, I've been writing application_server.

flexx/app/examples/flexx_in_thread.py
almarklein: In this particular case you can, because at that moment there won't already be a server. But it would break if somehow a server is created already. Therefore I think its better to include the call to `app.create_server()`, to make it more rubust/portable.

almarklein:
>  Maybe "application" is a bit more accurate. 

But that results in 
```python
from flexx import app
app = app.get_current_app()  # grrr, confusion
```

Korijn:
@almarklein I think the best thing you can do to avoid confusion is to mirror tornado conventions... so only call something an "application" when it is actually a `tornado.web.Application` and only call something a "server" when it is a `tornado.tcpserver.TCPServer`.

almarklein:
@Korijn So `app.create_xxx()` and `app.get_current_xxx()`, but what name for `xxx` that sounds app-like, server-like, loop-like?

Korijn:
Originally I thought you already returned tornado objects, and only just now did I notice that you have an abstraction layer in place. So I thought it was already appropriately named. :) But I guess this complicates things a bit.

Do you really still have the flexibility (ha) to move to another framework than Tornado though? It seems like Flexx is pretty tightly coupled at this point.

almarklein:
Right, I meant the abstraction object, which combines a loop, server and app.

This PR actually made the coupling less tight, because new server objects can be created (which could be based on a different backend). Supporting anything other than Tornado is not on my direct wishlist, but I'd like to keep the option open.

Korijn:
In that case, I think you've made the right decision here.

almarklein:
Ok, `app.create_app()` sucks because `app.create_app().app` looks terrible. As it is now, `app.create_server().server` is not great either. I also don't quite like `app.create_proxy()`. In lack of anything better, I am leaving this as it is, unless someone comes up with a good name/solution.

Korijn:
Well, currently, we're struggling more with the fact that you *must* go through `create_app()` to get to the native `app` object. We want to configure it (by adding a custom request handler) before launching a thread that actually starts. But if we move the call to `create_server` into the thread, we have no reference to app anymore... drama ensues :)

PR #154, opened 2016-07-05
Add create_server(new_loop=True), and rename current_server

almarklein:
Follow-up on #153 

This makes it possible to do:
```python
server = app.create_server(new_loop=True)
# do stuff with server.serving and server.app
t = threading.Thread(target=app.start)
t.start()
```

Also renamed `app.get_current_thread()` -> `app.current_thread()`

cc @korijn @Maxyme

### DIFF
--- a/flexx/app/tornadoserver.py
+++ b/flexx/app/tornadoserver.py
@@ -7,6 +7,7 @@ import json
 import time
 import socket
 import traceback
+import threading
 from urllib.parse import urlparse
 # from concurrent.futures import ThreadPoolExecutor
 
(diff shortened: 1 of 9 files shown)


flexx/app/tornadoserver.py
```diff
@@ -101,16 +107,25 @@ class TornadoServer(AbstractServer):
...
+    
     def _open(self, host, port):
         
-        # Get the current ioloop for the current thread
-        self._loop = tornado.ioloop.IOLoop.current()
+        # Get a new ioloop or the current ioloop for this thread
+        if self._new_loop:
+            self._loop = IOLoop()
+        else:
+            self._loop = IOLoop.current(is_main_thread())
```
Korijn: Kwarg missing?

almarklein: Not strictly necessary, but nicer, yes.

Issue #120, opened 2016-05-31
Should a handler also receive sub-events when using dynamism?


almarklein:
Consider:
```python
@connect('childen.*.mouse_down')
def _handle_child_clck(self, *events):
     for ev in events:
           ...
```
The handler will probably use `ev.source` to get the corresponding child and then do something with it. However, this handler will also get an event when the `children` property changes. Therefore it likely needs a `if ev.type == 'mouse_down'`, but this might be confusing.

On the other hand, it would make sense for this handler:
```python
@connect('childen.*.count')
def _update_total_count(self, *events):
    self.total_count = sum([child.count for chuld in children])
```

... which can probably use `@connect('children', 'childen.*.count')` in that case.

cc @Korijn

almarklein:
BTW: this behavior probably stems from the "reactive" model that used signals instead of events.

Korijn:
I think the distinction lies in the fact that we're subscribing to a property in one case (so we're probably defining a computed value over the set of properties) and we're listening to an interaction event in the other where the collection of emitters isn't relevant.

```python
@connect('children.*.count')
def _update_total_count(self, *events):
    self.total_count = sum([child.count for child in children])
```

If `children` changes, it makes sense that this handler is called, since the collection of `count` _properties_ we're subscribed to changes:

```python
@connect('children.*.mouse_down')
def _handle_child_clck(self, *events):
     for ev in events:
           ...
```

However, in this case, we're subscribed to a collection of `mouse_down` _event emitters_. Our handler is not a "computed value", but an interaction handler. We don't care about the collection of event handlers, we only care about the events as they are raised.

almarklein:
Yes, but changing the behavior based on whether the event corresponds to a property or not seems like it will lead to even more confusion.

I think I like the idea that you only get events for the latest element in a 'x.y.actual_event'. It's easy and explicit to add an extra event to `@connect()`. The downside is that in a case like calculating the total count, one might forget to do that. My feeling is that the advantages outweigh this downside though. WDYT?

Korijn:
Explicit > implicit, so I would agree with your preference in this case.

Korijn:
Thinking about this some more; changing a collection of event emitters like `mouse_down` doesn't alter the aggregated set of values, changing a collection of properties does. Is that really so counter-intuitive? I'm trying to think of a use case where changing the behaviour for props-or-not would actually e confusing.

almarklein:
[...]

PR #195, opened 2016-09-01
Improvements to event and property system

### LINKED ISSUE #146: Initial event for properties is only send to handlers of same object
```python

from flexx import event

class Foo1(event.HasEvents):
    
    @event.prop
    def bar(self, v=0):
        return v
    
    @event.connect('bar')
    def on_bar(self, *events):
        print('Foo1 saw bar change')
    
class Foo2(event.HasEvents):
    
    def __init__(self):
        self.x = Foo1()
        super().__init__()
    
    @event.connect('x.bar')
    def on_bar(self, *events):
        print('Foo2 saw bar change')

>>> f = Foo2()
Foo1 saw bar change
>>> f.x.bar = 3
Foo1 saw bar change
Foo2 saw bar change
```

This was initially somewhat intended. Or at least a consequence of how the initialization worked, which at the time seemed acceptable. But I just ran into this and banged my head into the wall, so I now consider it a bug.

The point is that each object sends initial events at the end of initialization. But `x` does that at a time when the handlers of `f` are not yet connected. Maybe it makes sense to have each object check all handlers and create initial events for itself when such a handler is attached to a property.

### LINKED ISSUE #172: Variable-depth event connection strings
_This issue is mostly meant to get some discussion going. It could very well turn out that I completely misunderstood some aspect of the current event system that already supports all this._

Imagine a `ComplicatedWidget` that contains a collection of widgets in an evented property `children`. Each of those widgets also contains a collection of widgets in an evented property `children`. Those widgets each have an attribute `slider` containing a `ui.Slider`.

Now, I want to subscribe to all the sliders that are in my `ComplicatedWidget`. Using dynamism, I understand from the docs that this is advertised to work as follows:

```python
class Example(ui.Widget):
    def init(self):
        self.cw = ComplicatedWidget()

    @event.connect('complicated_widget.children.*.children.*.slider.value')
    def on_slider_change(self, *events):
        pass
```

I don't like that I have to reach deep inside of `ComplicatedWidget` here. The point of that widget was to hide the complexity, so I can reuse it and change its internal composition without worrying about other code breaking. Obviously, if I decide to change the hierarchy, this code will break!

When I presented this to a colleague, they asked if I couldn't subscribe to the slider event by letting it bubble up instead, similar to how JS events traverse the DOM. Here's an example of what I would imagine this to look like:

```python
class Example(ui.Widget):
    def init(self):
        self.cw = ComplicatedWidget()

    @event.connect('slider.value')  # this bubbles up the hierarchy and is caught here
    def on_slider_change(self, *events):
        pass
```

You can make it just as flexible as JS event bubbling by exposing `stopPropagation()` and `preventDefault()` on the event object:

```python
    @event.connect('slider.value')
    def on_slider_change_2(self, *events):
        for ev in events:
            ev.stopPropagation()  # don't let the event bubble up any further
            ev.preventDefault()  # prevent the default handler of this event from running (which would be the update of the slider itself?)
```

A second weakness is that by subscribing to the event using the attribute name, I also risk code breakage if the evented `slider` property is renamed. Maybe subscribing to the event by source would be even more robust? E.g.:

```python
class Example(ui.Widget):
    def init(self):
        self.cw = ComplicatedWidget()

    @event.connect('value', source=ui.Slider)
    def on_slider_change(self, *events):
        pass
```

(this bears some similarity to Django receivers: https://docs.djangoproject.com/es/1.10/topics/signals/#connecting-to-signals-sent-by-specific-senders)

### LINKED ISSUE #176: More helpful error message when accidentally overwriting reserved properties like self.container
```python
class Foo(ui.Widget):
    def init(self):
        with ui.VBox() as self.container:
            # remainder of widget construction here
```

This seems like a natural name for the "container" of the subwidgets, but surprise: this will break your application, because `self.container` is already used internally. This could use a more helpful error message than `Host is not attached.`!

### LINKED ISSUE #192: Document init() and event handling cycle
It's hard to figure out in which order event.properties are set, the init() function is called, events are emitted, and handlers are executed. Some examples and a thorough explanation would be very welcome!

almarklein:
* A quick fix (fixes #176).
* When a handler connects to an event that corresponds to a prop, it receives an initial event (fixes #146).
* Subsequently, with "children*.value", "value events will be emitted when children changes.
* Changed list connection syntax  to "children*.value" (dropped the dot) old syntax still works though.
* Added ability to connect to a deep event via "children**.value" (fixes #172).
* Documented in docstring and by example the order of initialization (fixes #192)
* Implemented `HasEvents.__del__` to call `dispose()`.
* Fix issue that recursive pyscript std dependencies were not found.
* Generate more efficient JS by replacing `prototype` by Greek capital P.

### DIFF
--- a/flexx/app/examples/init_order.py
+++ b/flexx/app/examples/init_order.py
@@ -1,0 +1,47 @@
+"""
+This example demonstrates the orde of initialization:
+
+* Initial property values are set
+* The init() method gets called
+* Events are handled.
+"""
+
+from flexx import app, event
+
+
+class Example(app.Model):
+    
+    def init(self):
+        print('Py: in init')
+        self.spam = 5
+    
+    @event.prop
+    def foo(self, v=2):
+        print('Py: setting foo')
+        return v
+    
+    @event.connect('foo')
+    def on_foo(self, *events):
+        for ev in events:
+            print('Py: handling %s event' % ev.type, self.foo + self.spam)
+
+    class JS:
+        
+        def init(self):
+            print('JS: in init')
+            self.eggs = 5
+        
+        @event.prop
+        def bar(self, v=2):
+            print('JS: setting bar')
+            return v
+        
+        @event.connect('bar')
+        def on_bar(self, *events):
+            for ev in events:
+                print('JS: handling %s event' % ev.type, self.bar + self.eggs)
+
+
+if __name__ == '__main__':
+    m = app.launch(Example)
+    app.run()
--- a/flexx/ui/examples/deep_event_connections.py
+++ b/flexx/ui/examples/deep_event_connections.py
@@ -1,0 +1,37 @@
+# doc-export: DeepEventConnections
+"""
+Example that demonstartes how one can connect to events deep in the
+hierachy. Instead of using the star notation to select all children,
+one can use a double-star to select also the children's children, and
+their children, etc.
+"""
+
+from flexx import app, event, ui
+
+
+class DeepEventConnections(ui.Widget):
+    
+    def init(self):
+        # Put a label and some sliders deep in the hierarchy
+        
+        with ui.HBox():
+            with ui.VBox(flex=1) as self.box:
+                self.label = ui.Label()
+                for i in range(5):
+                    ui.Slider(value=i/5)
+    
+    class JS:
+        
+        @event.connect('children**.value')
+        def on_slider_change(self, *events):
+            # Show sum of slider values in the label
+            v = 0
+            for s in self.box.children:
+                if hasattr(s, 'value'):
+                    v += s.value
+            self.label.text = 'Sum of slider values is ' + str(v)
+
+
+if __name__ == '__main__':
+    m = app.launch(DeepEventConnections)
+    app.run()
--- a/flexx/app/examples/init_order1.py
+++ b/flexx/app/examples/init_order1.py
@@ -1,8 +1,8 @@
 """
-This example demonstrates the orde of initialization:
+This example demonstrates the order of initialization:
 
-* Initial property values are set
-* The init() method gets called
+* Initial property values are set.
+* The init() method gets called.
 * Events are handled.
 """
 
--- a/flexx/app/examples/init_order2.py
+++ b/flexx/app/examples/init_order2.py
@@ -19,7 +19,8 @@ class SubModel(app.Model):
     @event.connect('foo')
     def on_foo(self, *events):
         for ev in events:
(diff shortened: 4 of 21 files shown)


flexx/app/examples/init_order.py
```diff
@@ -0,0 +1,47 @@
...
+        
+        @event.connect('bar')
+        def on_bar(self, *events):
+            for ev in events:
+                print('JS: handling %s event' % ev.type, self.bar + self.eggs)
+
+
+if __name__ == '__main__':
+    m = app.launch(Example)
+    app.run()
```
Korijn: This demonstrates the order of things for a single widget. What happens to the order if you put this widget inside another widget that also has props, init() and event handlers, (and another one ...)?

almarklein: All the initialization of that subwidget happens at the moment when it is created. I added an example for a nested case too.

flexx/ui/examples/deep_event_connections.py
```diff
@@ -0,0 +1,37 @@
...
+    
+    class JS:
+        
+        @event.connect('children**.value')
+        def on_slider_change(self, *events):
+            # Show sum of slider values in the label
+            v = 0
+            for s in self.box.children:
+                if hasattr(s, 'value'):
+                    v += s.value
```
Korijn: Could you also write `v = sum(e.new_value for e in events)`?

ivoflipse: `new_value` isn't the increment, but the actual new value. But yes, doing over `s.value` should be possible

almarklein: Yes exactly. The problem with #120 was that a handler like this would receive "children" events. That (still) does not happen anymore. You only get events of the type corresponding to the tails of the connection strings.

Korijn:
> Generate more efficient JS by replacing `prototype` by Greek capital P.

Well, I don't know what to say! I'd like to know how that works!

almarklein:
Haha, yeah its a bit weird, and maybe we should change that to something better. The idea is that `SomeClass.prototype` is used a lot, making the name shorter saves space. The size of the JS is actually a bit of an issue on Windows; when I added new functionality to HasEvents (like the deep connection) the JS increases, and at some point it wont be possible to send the code via the `node` CLI.

I used a non-ascii char to reduce the chance of name clashes. It was also a nice exercise to make Unicode work on py2k. But other than that it's rather ugly.

Korijn:
You might also want to consider minification combined with a source map :)

Issue #194, opened 2016-09-01
Connection string micro language


almarklein:
The connection string is already a powerful mechanism to connect to events. With dynamism, selection of all elements in a list, and in #172 deep selectors. Let's think of what we want the language to look like. How far do we want to take it. Are there any other useful features we could add?

I think the language should be simple and easy to remember. I don't think making it look like regexp is a good idea. Rather make it look more like Python. 
- We have paths / dynamism: `parent.title`, `foo.bar.spam.eggs`
- We have selecting all: was `children.*.value` -> new syntax per #172 is `children*.value`, could possibly also be `children[:].value`?
- In that line of thought we couls also allow `children[0].value` and `children[:-2].value`. I can hardly imagine a use-case though.
- Deep selectors: `children**.value`
- Optional selectors? `children.children?.value` looks either one or two levels deep.
- Force connection: if per #188 we will raise an error when connecting to an unknown event, `foo!` can be used to connect anyway.

cc @korijn @ivoflipse

Korijn:
* Optional selectors are already supported, no? `children*.value, children.children*.value` is equivalent, I think.
* Force, as in, simply ignore the would-be error? -> Seems like an excellent idea. I would suggest putting the symbol at the start of the connection string though, rather than somewhere in the middle. E.g. `!foo.bar`.
* The use case for numerical selection of children will probably come around at some point. I like the syntax, anyhow.

Since we're in a micro language here, I think you should only use python-ish syntax when it's valid python syntax, to avoid confusion and bad assumptions. So, this is how I would expect it to work:

* `children*.value`
* `children**.value`
* `children[-1].value`
* `children[0].value`
* `children[:-1]*.value` (you always have to "unpack" collections, to be consistent)
* `children[:]**.value` -> this is equivalent to just `children**.value`

almarklein:
> you always have to "unpack" collections, to be consistent

I'd say that if we go with a Python-looking `children[:].value`,  that the star is not needed. I kinda like that, but then how would we then write the deep selector? Maybe `deep(children[:]).value`. This is quickly starting to feel too complex IMO. It is a micro language, but maybe having it *not* look like Python is a virtue.

I suspect that optional selectors are not worth the effort; you can just write two connection strings in such cases. I have doubts about the usefulness of the index/slice selectors too. 

Let me phrase some options:

* A
  * `children[:].value` selects all children
  * `children[0].value` selects first child
  * `children[1:].value` selects all but first child
[...]

PR #208, opened 2016-09-14
Warn if handler connects to event that is not known

### LINKED ISSUE #188: event.connect to non-existent event: no error?

almarklein:
Fixes #188 

There are two options when an event is connected that is not known:
* warn (as we do now)
* error

I think a warning suffices and is more backward compatible.

To do:

* [x] Decide what syntax do we use to suppress the warning. It's now done by prepending the connection string with an exclamation mark (related to #194).
* [x] docs

cc @korijn


Korijn:
I'm cool with that. I think some languages use `@` to suppress errors, but eh, if it's unambiguous and clearly documented, it should be good!

PR #217, opened 2016-09-18
Better prevention of context menu (mostly Firefox)

almarklein:


almarklein:
Also noting here that on Chrome, if you hold shift, the scroll becomes a horizontal scroll. In apps that don't care about the direction of scrolling you can simply do `scroll = ev.vscroll or ev.hscroll`.

almarklein:
This breaks checkboxes and radiobuttons (because they need that default action ... PR coming.

PR #219, opened 2016-09-20
Update (use of) Phosphor

### LINKED ISSUE #181: TabPanel: current tab property
The TabPanel currently lacks programmatic controls. E.g. a `current_tab` property setter with event handling would make it much more useful.

almarklein:
* [x] Update Phosphor-all packager to make use of new Phosphor mono-package
* [x] Make Flexx work with latest Phopshor version
* [x] Make Flexx work *better* with latest Phopshor version. E.g. dispose phosphor widgets, `TabPanel.current` (fixes #181).
* [ ] ~~Allow Flexx to wrap Phosphor widgets at runtime~~ this is real hard, needs more thought --- and an actual use case --- first
* [x] Update CSS for specific layouts
* [x] Apply "desktop-mode" better (without changing css of html or body elements?)
* [ ] ~~Revive GridPanel (cause Phosphor now has it)~~
* [ ] Make Flexx use `require()` or something else that better aligns with Jupyterlab et al.
* [ ] Make it possible for Flexx to *not* use it own shipped Phosphor, but the Phosphor that's already loaded in the browser.


Korijn:
What loader does jupyterlab use, anyway? Webpack?

almarklein:
> Webpack?
Yes

almarklein:
Remaining tasks can be done in a new PR.

PR #220, opened 2016-09-21
Fix doc building

almarklein:
* [x] Our CI should test that the docs build correctly, but this is not the case
* [x] Our docs currently don't build


almarklein:
The issue was revealed with the `with_bokeh` example. Because that example does not run on CI, the problem in `session.py` was never revealed.

PR #221, opened 2016-09-21
Prevent default event actions only for canvas

### LINKED ISSUE #174: Stopping JS event propagation
We need a way to allow the user to influence whether certain events (mouse, scrolling, keyboard) are allowed to bubble up, or whether their default action should be prevented.

This is pretty much impossible to do by letting handlers call a method on the event object because:
* events are supposed to be plain objects that can be send to Python, so methods would have to be special attributes.
* more importantly, events are scheduled at the event handlers and called in a next event loop iteration, which means that you'd be too late.

I think we need something like a property on `Widget` that you can set per-widget to prevent propagation of certain events. Maybe a `w.prevent_event_default = True`, though maybe something that can be more specific (e.g. allow context menu, but not scrolling).

Kind of events we may want to "catch":
* Context menu
* Scrolling
* Key events (think F5 and other browser shortcuts etc.)
* Middle mouse button, to prevent anchor cursor
* Mouse down/up/move events ?  I don't think this is necessary, ~~because there are no "default actions", right?~~ There are. Context menu on Firefox, and checking a radio button or checkbox goes via "default action".

Context menu is already prevented, otherwise we cannot capture right mouse button clicks. The disadvantage is that you cannot open the browser context menu anymore, e.g. to show the source. We can be smarter about that.

almarklein:
Closes #174 

Based on the observation that preventing context menu etc. is mainly a concern for apps that make use of the canvas for user interaction, I've moved the preventdefault-logic to the canvas.

In theory someone could want extensive user interaction with html elements. in that case he/she can look at how its handled in the `Canvas` and apply that approach.

For now I've left open a way to invoke the context menu by holding down all 3 modifier keys. Let's see how that works out.

I did a bit of work on making capturing wheel events work in a nice way. I.e. allow the canvas to prevent page scrolling but not when the canvas just happens to "pass by"  during scrolling. I.e. for interactive apps in the notebook. Such apps should set the class attribute `CAPTURE_WHEEL` of the `Canvas` subclass to `True`. Its not tested that well, but it's probably a good start, anyway.

### DIFF
--- a/flexx/ui/_widget.py
+++ b/flexx/ui/_widget.py
@@ -482,15 +482,6 @@ class Widget(Model):
             self.node.addEventListener('keyup', self.key_up, 0)
             self.node.addEventListener('keypress', self.key_press, 0)
             
-            # Disable context menu so we can handle RMB clicks
-            # Firefox is particularly stuborn with Shift+RMB, and RMB dbl click
-            if not window.flexx._disabled_context_menu:
-                self._disabled_context_menu = True
-                _context_menu = lambda ev: ev.preventDefault() and False
-                window.document.addEventListener('contextmenu', _context_menu, 0)
-                window.document.addEventListener('click', _context_menu, 0)
-                #window.document.addEventListener('dblclick', _context_menu, 0)
-            
             # Implement mouse capturing. When a mouse is pressed down on
             # a widget, it "captures" the mouse, and will continue to receive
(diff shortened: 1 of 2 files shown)


almarklein:
Together with the fact that emitters can be overloaded (and `super` can be used in them), I think we have enough to close #174.

Issue #224, opened 2016-09-29
Use JS modules and support loading modules from requirejs and webpack


almarklein:
Main use case: enable loading Flexx in Jupyterlab. Most importantly, Flexx should be able to load the phosphor that's already loaded in the browser, which is a WebPack module. 

Flexx' implementation is implemented in #225.

The main idea as it is now:
- Flexx exports its JS modules in this format: `define(name, [some, deps], function(some, deps) { ... })` which is compatible with AMD (e.g. requirejs).
- Flexx implements `window.define` unless `window.define` already exists (e.g. in the notebook). 
- Flexx' version will be a simple mechanism to tie modules together, no async and probably only supporting the format shown above.
- There is no global `define` function in Webpack, so Flexx' `define()` should include some smartness to make this work.
- It might well be necessary to implement a JS plugin for Jupyterlab that provided some sort of hook for Flexx' `define()` function to use.

almarklein:
This is implemented in #271, though we do not attempt to hook into requirejs; Flexx always brings its own loader.

PR #225, opened 2016-09-29
Export our JS using AMD-compatible modules

almarklein:
This is related to #224 (fixes part of it). The description/discussion here focuses on the implementation in Flexx. Implements #231.

This PR turned rather epic. PR's merged into this: #254, #264, #266.

- [x] Improve reporting of PyScript transpilation errors.
- [x] Remove support for import syntax in PyScript.
- [x] PyScript has better (more explicit) support for modules (of different types).
- [x] Provide a mechanism to create JS assets as modules. Dependencies and exports can be defined on these module objects, and to some extend automatically detected.
- [x] Refactor asset manager to be aware ~~(or not?)~~ of modules.
- [ ] ~~Put JS loader logic in PyScript subpackage~~
- [x] Make sure that it works with requirejs (i.e. in the notebook)
- [x] Automate the dependency of PyScript std (and importing its names).
- [x] Make export work again; it breaks the docs.
- [x] Fix / add tests
- [x] Implement backward compatible `create_module_assets()` function for now.
- [x] Can get rid of `flexx.classes` prefix for Model classes.
- [ ] ~~Can get rid of mangling to make `isinstance()` (and `super()`?) work for Model classes.~~
- [x] Allow interactive mode (besides the notebook), so we can better test module loading.
- [x] Update docs on assets and data.
- [x] Update docs on PyScript modules.

Next
- [ ] Improve Flexx loader so it can load modules out of order.
- [ ] ~~Load remote assets without deps at the end.~~
- [ ] Make it work with ~~webpack~~ `window.define()` provided by Jupyterlab
- [ ] PyScript std can probably reduce/remove prefixes, since they are not global anymore. Also look at reducing size of single-class modules (half of the code is stdlib imports :/ ).
- [ ] Minification (light-weight) of the modules we produce
- [ ] Maybe remove support for nodejs in `flexx.app` and `flexx.webruntime`; only targeting browsers simplifies things a lot. If not, make sure it works (again).
- [ ] ~~require using `import`?~~ we now do this via imports in Python!


almarklein:
Closing in favor of #271

Issue #213, opened 2016-09-15
Add CSS classname property to widget


almarklein:
Proposal: sometimes it's convenient to be able to add a CSS classname to a widget. You can sort of do this by subclassing the widget (which will add `.flx-YourSubClass` to the list of class names), but sometimes you may want more control, e.g. to style stuff using a 3d party theme, or to allow access of elements via Selenium.

almarklein:
Closed by #227

PR #227, opened 2016-10-02
Widget.css_class_name

almarklein:
Implement #213 

The name "css_class_name" is a bit long, but I fear that "class_name" will confuse people. Maybe "css_name", or "style_name"? I'd also consider renaming "style" to sorta group these two props together.

### DIFF
--- a/flexx/ui/examples/mondriaan.py
+++ b/flexx/ui/examples/mondriaan.py
@@ -8,6 +8,7 @@ from flexx import app, ui
 
 
 class MyVBox(ui.BoxLayout):
+    
     def __init__(self, **kwargs):
         kwargs['spacing'] = kwargs.get('spacing', 15)
         kwargs['padding'] = 0
@@ -25,6 +26,11 @@ class Mondriaan(ui.Widget):
     
     CSS = """
     .flx-Mondriaan {background: #000;}
+    .flx-Mondriaan .edge {background:none;}
+    .flx-Mondriaan .white {background:#fff;}
+    .flx-Mondriaan .red {background:#f23;}
+    .flx-Mondriaan .blue {background:#249;}
+    .flx-Mondriaan .yellow {background:#ff7;}
     """
     
     def init(self):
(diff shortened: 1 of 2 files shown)


Korijn:
Do you think this might be a nice grouping?

* `my_widget.css.class_name`
* `my_widget.css.style`

Korijn:
Ironically though, they're just as much HTML attributes as some other Flexx props like `flex`.

almarklein:
I am not sure about the `css`  sub object. It would need another `Model` object to make work nice. The name(s) should be such that it does not confuse Pythonista's who don't know much about CSS/HTML, while web dev can guess that it might be related to the css class name.

E.g. `style_name`, `css_name`, `style_class` , `css_class`.

I think I like `style_class`

Korijn:
I see your point, however, `class_name` or `css_class_name` seems most preferable to me. It's a class name, after all.

almarklein:
I can see how adding the word "class" will help web devs a lot to see what its for. To prevent Pythonistas getting confused we should prefix it with `style` or `css`, IMO. The suffic `name` is not that important, I think, e.g. you also write `<div class='xx'>` and not `<div class_name='x'>`.

Korijn:
`css_class` then?

flexx/ui/examples/mondriaan.py
```diff
@@ -33,25 +39,25 @@ def init(self):
...
                 
                 with MyHBox(flex=2):
-                    ui.Widget(flex=6, style='background:#fff;')
+                    ui.Widget(flex=6, css_class_name='white')
                     
                     with MyVBox(flex=1):
-                        ui.Widget(flex=1, style='background:#fff;')
-                        ui.Widget(flex=1, style='background:#ff7;')
+                        ui.Widget(flex=1, css_class_name='white')
+                        ui.Widget(flex=1, css_class_name='yellow')
```
Korijn: This is the first time I've seen [presentational class names](https://seesparkbox.com/foundry/naming_css_stuff_is_really_hard) that are simultaneously also semantic. :)

Issue #214, opened 2016-09-15
Provide ScrollAreaWidget or explain how to do it with CSS


eudoxos:
I would like to have a VBox where all children have their natural height and the VBox is vertically scrollable (similar for HBox). It is the top-level widget (not sure if it makes a difference). Qt has QScrollArea, maybe this is achievable with CSS here, but not clear how. Whatever I try with style (e.g. `style='overflow-y:scroll;'), I can't scroll that box. See e.g. this MWE:

```
from flexx import app, ui, event, pyscript
class Vvv(ui.Widget):
    def init(self):
        with ui.VBox(style='overflow-y:scroll;'):
            for i in range(5):
                with ui.GroupWidget(title='Group #'+str(i)):
                    with ui.VBox():
                        for j in range(5):
                            print(i,j)
                            ui.Slider(value=.2*j)
app.launch(Vvv)
app.start()
```

There are 5 sliders in each group, but I am seeing this (depending on the size of the window), note how the slider are clipped vertically (centered within the group), and that the vertical scrollbar area _is_ in the window but probably deactivated by Phosphorus or something like that:
![vvv](https://cloud.githubusercontent.com/assets/1029876/18548170/ce46a86a-7b47-11e6-9723-d2bc54358d16.png)

PR #230, opened 2016-10-08
Functionality to create widgets representing plain HTML elements

### LINKED ISSUE #177: Widget for representing HTML elements (Div, etc.)
Having something like the [Bootstrap grid](http://getbootstrap.com/css/#grid) makes development of responsive layouts super-easy. I suppose this could be integrated into Flexx relatively easy by providing a property on Widgets with which you can set [flex-wrap](https://css-tricks.com/almanac/properties/f/flex-wrap/), or by providing properties with which you can set the breakpoint from which they hold (like Bootstrap's `col-md-2` etc.).

This will probably require some discussion! We could alternatively look into enabling the use of systems like Bootstrap, although I don't see a short path to get there...

almarklein:
fixes #177

### DIFF
--- a/flexx/ui/__init__.py
+++ b/flexx/ui/__init__.py
@@ -63,6 +63,10 @@ To create an actual app from a widget, there are three possibilities:
 
 To lean about the individual widgets, check the 
 :doc:`list of widget classes <api>`.
+
+Web developers may want to have a look at the :class:`Div class <flexx.ui.Div>`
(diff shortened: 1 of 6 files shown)


almarklein:
touches on #214

Korijn:
Does this element also support the css class property?

almarklein:
> Does this element also support the css class property?

It's a widget, so yes, it does.

PR #232, opened 2016-10-12
Better deletion prevention of Model objects

### LINKED ISSUE #141: Generic way to avoid object deletion for properties that are a list of Models
Properties on `class Both` that are a list of Model instances, need some extra care to avoid deletion of the model objects. The reason is that "jitter" of the property when it is set from both Python and JS (e.g. during initialization) causes the property to fluctuate, making the objects go out of scope, being cleaned up by the GC, and subsequently deleted in JS as well.

One example of such a case is the `children` and `parent` property of `ui.Widget`.

The current solution is to explicitly keep the objects alive for a set amount of time when the property is set, using a simple "liveKeeper" object. However, there are other cases where such an approach would be needed, and having to implement such a thing feels hacky.

Ideally, Flexx would be able to handle this in an automatic way. If that is unfeasible, perhaps there is a more streamlined approach to help circumvent this problem.

almarklein:
Fixes #141 (good enough)

* Adds a `keep_alive()` method to ~~`Model`~~ `Session`, which keeps an object alive for a period of time that is expressed in ping-roundtrips. This means that the object will always be around long enough to survive synchronisation jitter, even if Python or JS are busy for several seconds (or hours, really).
* The `keep_alive()` method is called upon Model creation, which means users won't have to think about it in a great deal of cases.
* This also improves the ping-pong mechanism to determine connection timeouts. You should now not get these "closed due to pong" errors if you do heavy lifting in Python.
* This adds a secondary ping-pong mechanism to make `keep_alive()` work. One that is slightly higher level, so that it stalls when JS is doing work (which is what we want for the keep-alive use-case).

### DIFF
--- a/flexx/app/clientcore.py
+++ b/flexx/app/clientcore.py
@@ -185,11 +185,13 @@ class FlexxJS:
     def command(self, msg):
         """ Execute a command received from the server.
         """
-        if msg.startswith('PRINT '):
+        if msg.startswith('PING '):
+            self.ws.send('PONG ' + msg[5:])
+        elif msg.startswith('PRINT '):
             window.console.ori_log(msg[6:])
         elif msg.startswith('EVAL '):
             window._ = eval(msg[5:])
-            window.flexx.ws.send('RET ' + window._)  # send back result
+            self.ws.send('RET ' + window._)  # send back result
         elif msg.startswith('EXEC '):
             eval(msg[5:])  # like eval, but do not return result
         elif msg.startswith('DEFINE-JS '):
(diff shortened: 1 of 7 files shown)


Korijn:
LGTM!

almarklein:
Oh crap ... this is all wrong. There can be multiple sessions, each with their own ping counter, so the models should be stored per-session (yet global).

almarklein:
Ok, fixed. Also moved the `keep_alive()` method to `Session`, and can now be used to keep *any* kind of object alive. Also added tests. Merging.

Issue #140, opened 2016-06-21
Allow passing initial properties to app.serve and app.launch


almarklein:
```python
class MyObject(app.Model):

    @event.prop
    def foo(self, v=0):
        return float(v)

m = app.launch(MyObject, props=dict(foo=3))
```

Or maybe even `app.launch(MyObject, foo=3)`

cc @maxyme

Korijn:
Just a minor suggestion, but if you're adjusting this anyway, might be nice to also allow URL override, so e.g. an app can listen on the root.

almarklein:
@Korijn  do you mean like `app.start(hostname, port)`?  That's implemented on start since `app.serve` does not start the server, but only tells the server what apps to serve.

Korijn:
Here's an example:

```python
app.serve(r"admin/*", MyAdminApp)  # served at http://hostname.ext/admin/
app.serve(MyUnimportantApp)  # served at http://hostname.ext/MyUnimportantApp/
app.serve(r"*", MyMainApp)  # served at http://hostname.ext/
```

And as far as arguments go, this is what I understood @maxyme suggested:

```python
class MyObject(app.Model):

    def init(self, foo=0):
        self.foo = foo

m = app.serve(MyObject, dict(foo=3))
```

But your approach based on properties also seems viable and is more in touch with Flexx!

I'm not sure how that would work in scenarios where you're, for example, only passing in information that is subsequently used to _initialize_ a property though:

```python
class MyObject(app.Model):

    def init(self, secret_sauce=None):
        self.foo = Foo(secret_sauce)

m = app.serve(MyObject, dict(secret_sauce="magic"))
```

almarklein:
Aha, that makes sense. I was planning on allowing hosting apps on `http://hostname.ext` by e.g. naming an app `Index` or `Main`. But your suggestion makes much more sense.

I think I like the init being argumentless. The use-case that you described can in most cases be fixed by a smarter property setter (e.g. if not `isinstance(foo, Foo): ...`).

Maxyme:
What @Korijn wrote is in line with what I was suggesting. I have a small suggestion, in line with the tornado request helper class: changing init for initialize, then there is no more confusion as to what happens in the init since it's in the parent class.
```
class MyObject(app.Model):

    def initialize(self, foo=0):
        self.foo = foo

m = app.serve(MyObject, dict(foo=3))
```

almarklein:
Proposal:
[...]

PR #236, opened 2016-10-16
Model.init() can now have args

### LINKED ISSUE #228: Provide models with initial data
Sometimes a model needs information at initialization time, that you want to provide in a flexible way, something like:

```py
class MyModel(app.Model):

    def init(self, the_info):
         ...

...

m = MyModel(the_info=...)
```

Currently, the `init()` is always argumentless. The required behavior can be achieved by making a property `info`, the value of which can be used inside the `init()`. However, a property suggests that the info can be modified/reset after the model is instantiated, which is not always the case.

In other words, there are situations where a `Model` needs information at initialization time, but for which a property does not make sense conceptually.

Some possible solutions:

A. Any keyword arguments used to instantiate a `Model` which do not correspond to properties are passed to the `init()` as kwargs. You'd thus need to write it like `def init(self, info=None)` to make it a keyword arg. In theory it could also map to positional args by inspecting the signature, but that feels a bit too magical, I think. 

I am a bit worried about error reporting if a user makes a type in a property name. We would need to inspect the `init()` sig to see what kwargs it accepts, so we can be sure that a kwarg is intended for the `init()`. This should thus be disallowed: `def init(self, **kwargs)`.

B. Provide a new kind of property, similar to the current `readonly` that *must* be provided at initialization and is readonly afterwards. This adds less conceptual stuff for the user to learn, as it's just a kind of property.

C. Use positional arguments. The `Model.__init__()` could catch the positional arguments and pass them to the `init()`. The only downside is that for Models run as an app via `app.launch` or `app.serve` we'd need to provide a way to set these args, but we already do that for the kwargs, so I guess that could work ...

cc @korijn @Maxyme

almarklein:
Fixes #228. Relates to closed issue #140 and PR #199.

So you can do:
```py

class MyModel(app.Model):
    def init(foo, bar=3):
        ...

...
m = MyModel(7, 8, **properties_are_passed_as_kwargs_as_usual)
```

### DIFF
--- a/flexx/app/model.py
+++ b/flexx/app/model.py
@@ -331,7 +331,7 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
     # CSS for this class (no css in the base class)
     CSS = ""
     
-    def __init__(self, session=None, is_app=False, **kwargs):
+    def __init__(self, *init_args, session=None, is_app=False, **kwargs):
         
         # Param "is_app" is not used, but we "take" the argument so it
         # is not mistaken for a property value.
@@ -394,7 +394,7 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
         # subwidgets etc. This is done here, at the point where the
         # properties are initialized, but the handlers not yet.
         with self:
-            self.init()
+            self.init(*init_args)
         self._session._exec('flexx.instances.%s.init();' % self._id)
         
         # Initialize handlers for Python and for JS. Done after init()
--- a/flexx/app/model.py
+++ b/flexx/app/model.py
@@ -331,7 +331,12 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
     # CSS for this class (no css in the base class)
     CSS = ""
     
-    def __init__(self, *init_args, session=None, is_app=False, **kwargs):
+    def __init__(self, *init_args, **kwargs):
+        
+        # Pop args that we need from the kwargs (because legacy Python does
+        # not support keyword args after *args).
+        session = kwargs.pop('session', None)
+        is_app = kwargs.pop('is_app', None)
         
         # Param "is_app" is not used, but we "take" the argument so it
         # is not mistaken for a property value.
--- a/flexx/app/model.py
+++ b/flexx/app/model.py
@@ -336,10 +336,7 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
         # Pop args that we need from the kwargs (because legacy Python does
         # not support keyword args after *args).
         session = kwargs.pop('session', None)
-        is_app = kwargs.pop('is_app', None)
-        
-        # Param "is_app" is not used, but we "take" the argument so it
-        # is not mistaken for a property value.
+        kwargs.pop('is_app', None)  # not used here, but need to "take" it 
         
         # Set id and register this instance
         Model._counter += 1
--- a/flexx/app/tests/test_model.py
+++ b/flexx/app/tests/test_model.py
@@ -85,6 +85,13 @@ class Foo6(Foo1):
             return {}
 
 
+class Foo7(Model):
+    
+    def init(self, foo, bar=0):
+        self.foo = foo
+        self.bar = bar
+
+
 def test_pairing1():
     
     assert isinstance(Foo1.title, event._emitters.Property)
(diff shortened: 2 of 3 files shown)


almarklein:
One thing that I'd like to solve is how to make this work with `app.launch` and `app.serve`. API, wise, I think this would be really nice (and better than the current properties arg):
```py
app.launch(SomeModel.pre_init(3, 4, text='bla'), 'firefox')
```

Implementation wise, this needs some thought. Making `pre_init` return an actual `Model` class would make it Just Work, but would probably involve meta class stuff like `__new__`. It would have to be a class with the same name, but that's ... weird, and also all Model classes get registered to make them available in JS, which complicates things.

Easier is to wrap the class + args + kwargs in a simple object, and then make `app.launch` and friends aware of that.

Korijn:
With regards to `app.launch` and `app.serve`, I would think you can already pass initial data along as per https://github.com/zoofIO/flexx/issues/140?

almarklein:
That was only for keyword arguments. We could extend that approach for positional arguments, in which case it would look like:
```py
app.launch(SomeModel, 'firefox', init_args=[3, 4], properties={'text': 'bla'})
```
This should work, but I don't like it a lot, since I'd rather define the args and kwargs using an actual *call* to something. Or at least explore that idea :)

Let's try:
```py
# A: Extending current approach
app.launch(SomeModel, 'firefox', init_args=[3, 4], properties={'text': 'bla'})

# B: Using a factory func. Possibly useful in other ways. But rather verbose.
app.launch(SomeModel, 'firefox', factory=lambda: SomeModel(3, 4, text='bla'))

# C: Collecting args separately. Feels weird?
app.launch(SomeModel, 'firefox', model_args=app.get_args(3, 4, text='bla'))

# D: Collecting args and class together. I like the compactness.
app.launch(SomeModel.pre_init(3, 4, text='bla'), 'firefox')

# E: Like D, but more explicit.
app.launch(app.pre_init(SomeModel, 3, 4, text='bla'), 'firefox')
```

In `D` and `E`, the first object in `app.launch` is a "factory object" that exposes the class that it will create an instance of. Potentially, such objects can generate instances in other ways, bringing the same flexibility as `B`.

Thoughs, ideas?

Korijn:
What devs will try intuitively is:

```python
app.launch(SomeModel(3, 4, text='bla'), 'firefox')
```

:)

almarklein:
Cannot work; it needs a session. Also `app.serve(SomeModel(3, 4, text='bla'), 'firefox')` cannot work.

Korijn:
A seems easiest to me; kinda reminds of threading.Thread which should be familiar to most pythonistas!

almarklein:
> I would call 'em just args and kwargs though.

Mmm, maybe they should just be the `*args` and `**kwargs`. The thing holding me back with this is the fact that `app.export` and `app.launch` need to be able to receive additional info. But maybe this can be done by popping a name off the kwargs:

```py
# F: use args and kwargs of launch/serve
app.launch(SomeModel, 'firefox', 3, 4, text='bla')
# and export and launch would have 
app.launch(SomeModel, 'firefox', 3, 4, text='bla', runtime_options=dict(title='bla', size=(100, 100)))
app.export(SomeModel, 'firefox', 3, 4, text='bla', export_options=dict(inline=True))
```

almarklein:
(last attempt, otherwise we just go with A)

Korijn:
Here's an out of the box suggestion:

```python
app = App(SomeModel, 3, 4, text='bla')
app.launch('firefox', dict(title='bla', size=(100, 100)))
app.export(dict(inline=True))
```
Ultimately it's your decision, I can live with any option, but I think simplicity is key in getting devs to pick up on it.

Korijn:
Or perhaps:

```python
my_app = app.create(SomeModel, 3, 4, text='bla')
my_app.launch('firefox', dict(title='bla', size=(100, 100)
my_app.export(dict(inline=True))
```

Although, I find the `App(...)` constructor more intuitive since you are actually constructing an app. The global `app.create(...)` is less obvious; using an invisible global app to create sub apps. It's more like a server then, e.g. `server.create_app(...)`

almarklein:
I think I like that! Technically, its very similar to what I proposed with D and E, but by calling it an app and giving it some extra methods it makes a lot more sense.

So these would be equivalent:
```py
app.App(SomeModel, 3, 4, text='bla').launch('firefox', ...)
app.launch(app.App(SomeModel, 3, 4, text='bla'), 'firefox', ...)
```

almarklein:
Ok. I', merging the first bit, and will do the thing with the `App` class in #238, because I foresee some merge conflicts if I do it now.

PR #239, opened 2016-10-19
TreeWidget

almarklein:
I am *very* happy with this :)

* TreeWidget can contain TreeItems, which are Model objects (not widgets).
* TreeItems can contain sub items.
* Construct using context manager and manipulate using `treewidget.items` and `treeitem.items`.
* Items are selectable (via a policy set at the tree widget).
* Items can be made checkable.
* Items can be collapsed/expanded if they are said to be so, or if they have sub items.
* Is implemented almost entirely in CSS; all that the JS does is apply CSS classes.
* This makes it very easy to style. Even changing the checkbox glyph etc.

In combination with the event system and the deep connector (`event.connect('tree.items**.xxx')`) this work very nice, I think.

### DIFF
--- a/docs/scripts/genuiclasses.py
+++ b/docs/scripts/genuiclasses.py
@@ -24,7 +24,7 @@ def main():
         if isinstance(mod, ModuleType):
             classes = []
             for w in mod.__dict__.values():
-                if isinstance(w, type) and issubclass(w, ui.Widget):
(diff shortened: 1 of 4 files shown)


Korijn:
Awesome! Is there an example to check out?

almarklein:
https://dl.dropboxusercontent.com/u/1463853/tree.html

almarklein:
And here: http://flexx.readthedocs.io/en/latest/ui/tree.html

PR #241, opened 2016-10-21
Settle on a syntax for connection strings

### LINKED ISSUE #194: Connection string micro language

almarklein:
* Fixes #194: formalize the syntax for connection strings.
* Make sure normalization/checks are done in the right places.
* Fixes connection string that end in deep connector.
* Fixes a few other corner cases that where bugs but simply never showed up yet.
* The code grew somewhat, so I had to finally make `evaljs` pass the code to `node` via a file.
* Which gave some space, so I got rid of the `Ƥ` placeholder for `prototype`.
* Make functions not anonymous (which makes handlers for which no label is specified sill still have a sensible label).

### DIFF
--- a/flexx/app/_model.py
+++ b/flexx/app/_model.py
@@ -532,7 +532,7 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
             self._session._exec(cmd)
     
     def _register_handler(self, *args):
-        event_type = args[0].split(':')[0].strip('!')
+        event_type = args[0].split(':')[0]
         if not self.get_event_handlers(event_type):
             self.call_js('_new_event_type_hook("%s")' % event_type)
         return super()._register_handler(*args)
@@ -639,7 +639,7 @@ class Model(with_metaclass(ModelMeta, event.HasEvents)):
             window.flexx.instances[self._id] = 'disposed'
         
         def _register_handler(self, *args):
-            event_type = args[0].split(':')[0].strip('!')
+            event_type = args[0].split(':')[0]
             if not self.get_event_handlers(event_type):
                 self._new_event_type_hook(event_type)
             return super()._register_handler(*args)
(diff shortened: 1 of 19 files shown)


almarklein:
I'll wait with merging till Monday, just in case.

Korijn:
I think we had a case where the combination of ! and ** generated a failure, or at least an unintuitive error message.

almarklein:
That should be fixed by this PR (though now it should be written as `foo.children**!`; the old syntax (w.r.t. the position of the `!` will still work, but you'll get a warning.

almarklein:
... actually, now that I have slept over it, I think I'd like the `!` to be at the very start of the string, so `!foo.children**`. Som context: I initially thought that the ! could be applied to any part of the path, but this is not true; it only makes sense for the final step, so a connection string should be "forced" as a whole. Yesterday I thought the end would be the best place, but the reason was based on an implementation detail, the start is better, I now think, from the user's perspective.

PR #243, opened 2016-10-25
Fix tabindex and still make key events Just Work

almarklein:
This makes `Model` provide a method hook in JS that gets called when a handler is connected to an event that was previously unconnected. In `Widget` we use that hook to check if we need to support key events and set the `tab_index` if necessary. Over time, we might want to use this hook for other purposes as well (e.g. mouse events prevent-default stuff?)


Korijn:
Note that not all HTML elements actually support tabindex the same way. They sometimes even implement it differently.

almarklein:
Yes. This actually make Flexx be much more careful about setting `tabindex`. Previously the default value was `-1`, now the default value is `None` which means "do not set".

PR #245, opened 2016-10-26
Add combobox/dropdown

almarklein:
* A `ComboBox` is an optionally editable dropdown button
* A `DropdownContainer` can hide a widget out of sight.

### DIFF
--- a/flexx/ui/widgets/__init__.py
+++ b/flexx/ui/widgets/__init__.py
@@ -8,6 +8,7 @@ from .. import Widget
 from ._button import BaseButton, Button, ToggleButton, RadioButton, CheckBox
 from ._slider import Slider
 from ._tree import TreeWidget, TreeItem
+from ._dropdown import ComboBox, DropdownContainer
 from ._lineedit import LineEdit
 from ._label import Label
 from ._group import GroupWidget
--- a/flexx/ui/widgets/_dropdown.py
+++ b/flexx/ui/widgets/_dropdown.py
@@ -184,6 +184,8 @@ class ComboBox(BaseDropdown):
     with an editable text. It can be used to select among a set of
     options in a more compact manner than a TreeWidget would.
     Optionally, the text of the combobox can be edited.
+    Connect to the ``text`` and/or ``selected_index`` properties to keep
+    track of interactions.
     """
         
     CSS = """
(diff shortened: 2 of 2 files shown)


Korijn:
@ivoflipse

Korijn:
Can you provide different values for the option tag value and text attributes?

almarklein:
I am not sure what you mean. You can always set the `text` property programatically (to something that does not match an option).

The options are just strings, and these are the strings that are used to set `text`. If you want some sort of mapping, its probably easiest to use `selected_index`.

And not that I think of it, that should probably be renamed to `current_index`.

ivoflipse:
I think Korijn was referring to a select tag, where there's a difference between the value of an option and the text it displays in the dropdown

```
<select>
  <option value="volvo">Volvo</option>
  <option value="saab">Saab</option>
  <option value="mercedes">Mercedes</option>
  <option value="audi">Audi</option>
</select>
```

At the moment, the flexx widget doesn't support this behavior, since it's not a select tag, but a custom dropdown. We'd have to maintain our own mapping of human readable values to values our program would want to use

almarklein:
If it's a common use-case, we can think about supporting something like this. E.g. by allowing to specify tuples.

almarklein:
Though the `value` in `<option>` is sorta like the `index` in our implementation, except its a key instead of an int.

Korijn:
Our use case is a dictionary mapping abbreviations to full names going into the combobox as <option value="abbr">full</option>

almarklein:
What about adding `selected_key`,which contains the "key" corresponding to a selected item (`None` if no item is selected or if the selected item has no key). The `options` prop can then be a list of key-value pairs, or a just a dictionary.

A variant of this: rename `selected_index` to `selected_key`, or maybe just `selected`. This prop contains the key for the selected item, or its index if the selected item has no key.

Korijn:
`selected_index` together with dict or keyval pairs as `options` would be sufficient, I think.

almarklein:
Mmm ... that would mean you would lookup the key using `combobox.options[combobox.selected_index][0]`. Which seems rather non-trivial.

Korijn:
Right. So the selected key addition would make it easier.

Korijn:
Here's our use case:

image

We'd like to display human readable text, but still get the (currently displayed) value from a `selected_value` property.

So a dict would be much better.

almarklein:
See #250

PR #246, opened 2016-10-27
Fix dropdown hidden overflow

almarklein:
The expanded part of the dropdown is now `display: fixed`, so it is drawn over everything else.


almarklein:
On my machine the `DropdownContainer` looks strange in Edge. As if Edge does not invalidate the correct regions. Since that widget should be rare (the `ComboBox` looks fine), we can leave that as-is for now.

PR #250, opened 2016-11-01
ComboBox improvements: selected_key prop

almarklein:
Continuation of #245

* `ComboBox`s `options` property is now a list of tuples `(key, text)`.
* It also has `selected_key` analog to `selected_index`, but holding the key corresponding to the selected item (or None).

### DIFF
--- a/flexx/ui/widgets/_dropdown.py
+++ b/flexx/ui/widgets/_dropdown.py
@@ -70,13 +70,13 @@ class BaseDropdown(Widget):
             border: 1px solid black;
             min-height: 1.7em;
             max-height: 1.7em;
-            padding-right: 1.5em; /* space for the expand but */
+            white-space: nowrap; /* keep label and but on-line */
         }
         
         .flx-BaseDropdown > .flx-dd-edit {
             display: none;
             max-width: 2em;  /* reset silly lineedit sizing */
-            min-width: 100%;
+            min-width: calc(100% - 1.5em - 2px);
             min-height: 1em;
             margin: 0;
             padding: 0;
@@ -85,7 +85,7 @@ class BaseDropdown(Widget):
         
         .flx-BaseDropdown > .flx-dd-label {
             display: inline-block;
-            min-width: 100%;
+            min-width: calc(100% - 1.5em - 2px);
             min-height: 1em;
             user-select: none;
             -moz-user-select: none;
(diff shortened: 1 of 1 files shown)


almarklein:
I slightly dislike the fact that `options` being a tuple makes things seem more complex for simple cases. An alternative could be to add an `option_keys` property to hold the keys. Setting `options` with a dict would then auto-set that prop too. Though such an approach could more easily lead to inconsistencies (also in length) between both lists.

ivoflipse:
Perhaps `option_keys` should be used to lookup the value for the keys in the ComboBox. If something isn't in there, you just return None. Then hopefully it should be obvious if something is missing (unless people put None in as a value...) Plus, the ComboxBox doesn't break if there are inconsistencies. 

I guess the only benefit from this would be that it would safe the user from implementing such functionality every time they use the ComboBox

almarklein:
I think that if we introduce the concept of keys, we should definitely also have `selected_key`.

Maybe `option_keys` should be readonly; its automatically set when setting options. That way the two are sure to stay in sync.

Korijn:
Here's how it works in .NET:

ComboBox has all of the following:

* Items
* SelectedIndex
* SelectedItem
* SelectedText
* SelectedValue

All of them are getters and setters. But SelectedText and SelectedValue map to Items using SelectedIndex.

Items is a collection of objects (giving each item an Index), and each item has a Text and a Value property.

almarklein:
I looked at `option_keys`, also a version where they're hidden (they don't have much use other than using them internally to be able to set `selected_key`. But this is really hard to pull off because when JS/Py syncs the `options` property, it resets the `option_keys`, so this is a bad idea.

The .NET API seems to call things differently. Most importantly, it calls `values` what I called keys. In Qt they call this the "user data".

So what makes more sense:
```py
ComboBox(options={'foo': 'Foo Text', 'bar': 'Bar Text'})
```
or
```py
ComboBox(options={'Foo Text': 'foo', 'Bar Text': 'bar'})
```

Also, in the current situation, keys are forced to be strings. This could also be made to allow ints/floats.

Korijn:
Both are strings mapping to strings, so it doesn't matter to me. Keys are keys. :)

almarklein:
Ok, so the API is ok like this?

Korijn:
As the code in the PR is now, I'm happy.

PR #251, opened 2016-11-01
Prevent default action for key press if widget has listeners

almarklein:


Korijn:
How does this work? Tabindex is used to change the order in which you tab through all the UI elements, no?

almarklein:
Yes. But in this case it is only used as a marker; when `tabindex` is not `None` the widget is receptive to having focus and able to receive key events. There is code in this same module that sets `tabindex` to `-1` if a connection to a key event is made.

Korijn:
OK. Seems a bit brittle to me but I'm sure it'll work.

almarklein:
The worst that can happen is that the user sets the `tabindex` while there are no connections to key events. In that case the widget will prevent default actions even though it might not be necessary. A user setting `tabindex` to `None` would prevent receiving key events altogether.

Korijn:
The other worry I have is that the future us might not remember this a year from now, and break it when refactoring `tabindex` functionality. :)

almarklein:
Right. I get your point. But the `tabindex` code is in the same module and there is a clear comment why we use it, so I' relatively confident that this will work out ok. I owe you a beer if it does not :P

Korijn:
Deal.

almarklein:
Mmm, I just found that I cannot use F12 to open the developer console in a certain situation. Very annoying. Maybe we should tune this back a bit somehow...

almarklein:
Proposal:

A class attribute `CAPTURE_KEYS` that can be set to `True` to capture all, or to a function that can return True to capture only certain key combo's.

almarklein:
Another possibility, perhaps: 

The default actions occur on press. So we could allow the default event handlers for `key_down` to somehow set a flag to prevent the default action.

Korijn:
The hook where you can call preventDefault yourself seems most flexible to me.

almarklein:
#257

almarklein:
@Korijn thanks. You're right I was about to over-design it.

PR #252, opened 2016-11-03
Automatic assets - js modules match py modules

### LINKED ISSUE #244: How to write a JS module
In Python and JS to write some reusable functions you would put them in a separate file and just import that. How does this work in Flexx? How do I make a simple JS module? :)

almarklein:
Closes #244 

This PR is a continuation of #225. I wish I'd realized the potential of this earlier, but I guess I needed #225 first in order to see how we can take it further ...

This PR:
* Introduces the concept of a JS module (only used internally), that matches with a Python module.
* We can have JS modules for modules that define `Model` classes, or for fully PyScript compatible modules.
* PyScript now reports the globals (used, but undefined variables) of the transpiled code.
* We can use that information to detect what global variables of the Python module are used by JS.
* We can use that to collect dependencies and correctly import these names in the JS module
* Similarly we can see what functions/classes we need to transpile (and include the result in the JS module).

All this means that the user won't have to do any asset management, except for specifying external assets (e.g. Bokeh, JQuery, Phosphor), but this can likely be done by simply defining the asset in a module; we can detect it and automatically make that asset a dependency of the JS module.

This also means that we can *use* linting tools to evaluate our JS, instead of having to *fool* them.


Korijn:
Love it! A question: can you also use python libs in JS now? e.g. https://github.com/dgilland/pydash ?

almarklein:
Yes, I think you could! As long as a function/class is PyScript compatible you can import it in a module and use it in JS.

Korijn:
Awesome. Can't wait to see how this works out for our project. :)

almarklein:
Dear Git Gods, I made the stupid mistake of rebasing the branch on which this was based. Please be gentle on my soul ...

almarklein:
In other words: oh crap ...

Issue #253, opened 2016-11-04
Running in Jupyter notebook / lab / other


almarklein:
The purpose of this issue is to get an idea of what's needed for Flexx to run inside environments like the notebook, JupyterLab, and other. What special measures does Flexx need to take, and what does Flexx need to know. Some things are already in place, but there are some loose ends.

Flexx needs to be made aware of certain aspects of the environment that it runs in. Sometimes its ok if the user provides this information (e.g. via `app.init_notebook()`), but the more we can automate, the better. If the server is aware of something, its easy to make the client aware too. The reverse is less true (the server might need the information before a connection is made).

I split this up aspects related to notebooks and Phosphor.

### General

Since Flexx is embedded as part of an existing web page/app, there is no concept of a "Flexx app", and therefore must make use of the default session. E.g. `app.init_notebook()` takes care of this.

### Notebooks

Notebooks can run in:
* The Jupyer notebook (which does not have Phosphor loaded).
* The notebook widget in JLab (Phosphor is loaded).
* Something else?
* In all cases, its a notebook that can be *exported* to HTML. For this reason, Flexx must *always* embed its own `phosphor-all.js` and `phosphor-all.css` for the exported notebook to work.

What Flexx needs to be able to detect:
* Server and client need to know whether its running in a notebook. This is set by `app.init_notebook()`.
* Client needs to know whether its running in an *exported* notebook. We now try to detect this by looking for the presence of `window.IPython.foo.bar` and `window.jupyter.lab`. This seems fragile though.

### Phosphor

If Flexx is loaded in an environment that already uses Phosphor, Flexx should probably use that version. This is a problem we're now facing with integrating Flexx with JLab, but could be an issue in general, especially if Phosphor gets more widely adopted.

For the JavaScript, Flexx ships its own `phosphor-all.js` in an AMD module. In one way or another, we can make Flexx make use of the available library instead of its own. For the CSS, Flexx ships its own `phosphor-al.js` (i.e. `phosphor/styles/base.css`). Fortunately, we can set this style DOM element to `disabled` if we know that Phosphor is already loaded.

How to determine if Phosphor is already loaded:
* Check the presence of `window.jupyter.lab`. Works in JLab, but is not a generic solution.
* Phosphor could set a global (and also report its version), e.g. `window.phosphor_version`.

How to let Flexx reach the Phosphor lib:
* JLab (the Flexx-plugin) can provide hooks.
* A more generic solution?

[...]

PR #254, opened 2016-11-04
Making things work in Jupyterlab - part 1

### LINKED ISSUE #263: Modularize CSS
Flexx might not be alone on the page. While the JS is (soon) pretty modular, CSS is sort of global by nature. We don't want our CSS to mess up other parts of the page, or silly mistakes in other people's CSS to mess up our widgets.

* All our class names should be prefixed with `flx-`. Even if we would use strict selectors like `.flx-Widget > hidden`, our `hidden` class can easily cause trouble elsewhere.
* For clarity, modifier class names should be `flx-mod-x`.
* All third party class names (in particular Phosphor) should be "under a `flx-Widget`", e.g. `.flx-Widget p-TabBar`.

Further notes regarding Phosphor: the `phosphor-all.css` should consist of Phosphor's standard CSS plus any CSS that we need to add to make things work, and (as outlines above) should be "prefixed" with `.flx-Widget p-XX`. That way our CSS should not harm the rest of the page, and can also be disabled as a whole.

almarklein:
Relates to #253. Fixes #263.

This is mainly about CSS.

* [x] Make Flexx work (technically) in a notebook in JLab
* [x] Make Flexx not use its own Phosphor CSS when in JLAB
* [x] Make Flexx' styling of Phosphor stuff such that it does not clash
* [x] Put Flexx' Phosphor-related CSS in `phosphor-all.css` so it can be disabled as a whole
* [x] Prefix all Phosphor CSS with `.flx-Widget`.
* [x] Clean other CSS to avoid clashes.

By prefixing the CSS rules, we can already safely load our own `phosphor-all.css` in Jupyterlab without breaking things. But it would reset the style of all Flexx widgets, so we still keep the approach to disable `phosphor-all.css`.


almarklein:
#263 should take this further

PR #266, opened 2016-11-09
Automatic assets - js modules match py modules (2)

### LINKED ISSUE #244: How to write a JS module

almarklein:
Supersedes #252 (I broke that branch)

Closes #244 

This PR is a continuation of #225. I wish I'd realized the potential of this earlier, but I guess I needed #225 first in order to see how we can take it further ...

This PR:
* Introduces the concept of a JS module (only used internally), that matches with a Python module.
* We can have JS modules for modules that define `Model` classes, or for fully PyScript compatible modules.
* PyScript now reports the globals (used, but undefined variables) of the transpiled code.
* We can use that information to detect what global variables of the Python module are used by JS.
* We can use that to collect dependencies and correctly import these names in the JS module
* Similarly we can see what functions/classes we need to transpile (and include the result in the JS module).

All this means that the user won't have to do any asset management, except for specifying external assets (e.g. Bokeh, JQuery, Phosphor), but this can likely be done by simply defining the asset in a module; we can detect it and automatically make that asset a dependency of the JS module.

This also means that we can *use* linting tools to evaluate our JS, instead of having to *fool* them.


almarklein:
Some notes: in essense, the idea of modules is all about propagating dependencies:

* In PyScript we detect unresolved dependencies in JS code, and move these up
  the namespace stack.
* The create_js_hasevents_class() function and ModelMeta class collect the
  dependencies from the different code pieces.
* In JSModule we resolve some dependencies and let other propagate into
  module dependencies.
* In the Bundle class, again some dependencies are resolved due to bundling,
  and others propagate to dependencies between bundles.

PR #291, opened 2016-11-28
Allow running Phosphor provided by Jupyter

almarklein:
* Make use of lastest Phosphor-all, which has a new API that uses a `window.require_phosphor()` function.
* Flexx' module loader `window.flexx.define` + `window.flexx.require` detects when something `phosphor/...` is required, and calls `window.jupyter.require()` or `window.require_phosphor()`.
* All widget modules load needed phosphor modules using `window.flexx.require()` (made possible by `RawJS` #290).


almarklein:
cc @sccolbert @blink1073

Issue #303, opened 2016-12-04
serve apps through https


pigay:
Is it possible to configure the tornado server to serve through https?

It could be used with the [cloud webapp use case](Hosting-a-web-app-in-the-cloud]) for instance.

I tried a naive hack, but it seems that passing `ssl_options` to Tornado `HTTPServer` constructor is not enough: application page is frozen in starting state, while the server is polling somewhere in its even loop.

Is there a easy way to get the secure protocol working?

Thanks

almarklein:
Flexx currently does not provide a means to do this. It'd surely be possible but I don't know enough about this to say how hard it would be. If you could provide a proof of concept of sorts, I'm very open to include this functionality.

Apart from setting `ssl_options`, you'd probably also need to change some references of `http` to `https`.

almarklein:
And you'd have to do the same for the websocket connection.

pigay:
Here it is!

Cheers!

Pierre

almarklein:
Closed by #307 and #317

Issue #299, opened 2016-12-02
Q: code deployment on remote IPy notebook server


alexveden:
I've created new custom widget component, and it works fine on my development localhost machine. But when I try to deploy this code on remote IPy Notebooks server the widget is drawn, but it doesn't respond to mouse clicks.

I'm seeing this in browser console:
> Socket connected
> Lost connection with server: Could not launch app: RuntimeError("Asked for app id 'UugOTCOcFDkadMCunJqq8hyg', but could not find it",) (1003)

I suspect that the problem is in the settings, and I need to make the server to listen remote hostname.
My setup is:
flexx: 0.4.1
Ubuntu 16.04
IPy Notebooks server: https://10.0.1.2:8888/

I've tried to add custom config file to ~/.config/.flexx.cfg and add following line to it:
> config.hostname=10.0.1.2

but it still has no effect.

p.s. Could you clarify generic .flexx.cfg paths locations for different OSes inside documentation, because it's not so intuitive to find 'appdata' in ~/.config in Linux?

Thx

alexveden:
Changes, config contents to:
> hostname=10.0.1.2
and not it's connecting.

But another issue:
> The page at 'https://10.0.1.2:8888/notebooks/flexx_new.ipynb' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://10.0.1.2:49190/__default__/ws'. This request has been blocked; this endpoint must be available over WSS.

As I know there is no option to launch remote notebook server without HTTPS.

almarklein:
A lot has changed w.r.t. notebook support since 0.4.1, though the latest issue seems to suggest that non-secure websockets are blocked. I have not looked into this in any way, but I suspect that you can make Flexx use `wss` by editing `flexx/app/_clientcore.py`.

I've added a note to the docs on how to find the appdata dir: ``flexx.util.config.appdata_dir()``. Putting it in `~/.config` was a suggestion from a Linux user who said it was the best place to put it (previously I used `~/.flexx`).

almarklein:
I think the remaining problem is the issue of secure websockets, which is covered by #303, so closing this in favor of that issue.

PR #307, opened 2016-12-05
add support for HTTPS protocol

pigay:
Proof of concept for #303
* Allows to pass arguments to `tornado.HTTPServer` constructor through `app.create_server`.
* support for HTTPS protocol

Enable SSL transport by calling :
```
args = {}
args['ssl_options'] = {'certfile' : 'path-to-crtfile', 'keyfile' : 'path-to-keyfile'}
app.create_server(**args)

app.serve(Example, 'Example')
app.run()
```

### DIFF
--- a/flexx/app/_server.py
+++ b/flexx/app/_server.py
@@ -176,3 +176,9 @@ class AbstractServer:
         Or None if the server is not serving (anymore).
         """
         return self._serving
+
+    @property
+    def protocol(self):
+        """ Get a string representing served protocol
+        """
+        raise NotImplementedError
--- a/flexx/app/_server.py
+++ b/flexx/app/_server.py
@@ -50,6 +50,8 @@ def create_server(host=None, port=None, new_loop=False, backend='tornado',
             app.serve(Example, 'Example')
             app.run()
 
+        Alternately, cert and key files can be provided through
+           ``ssl_certfile`` and ``ssl_keyfile`` configuration variables.
     """
     # Lazy load tornado, so that we can use anything we want there without
     # preventing other parts of flexx.app from using *this* module.
--- a/flexx/app/_server.py
+++ b/flexx/app/_server.py
@@ -11,7 +11,8 @@ from .. import config
 _current_server = None
 
 
-def create_server(host=None, port=None, new_loop=False, backend='tornado', **server_kwargs):
+def create_server(host=None, port=None, new_loop=False, backend='tornado',
+                  **server_kwargs):
     """
     Create a new server object. This is automatically called; users generally
     don't need this, unless they want to explicitly specify host/port,
--- a/flexx/app/_server.py
+++ b/flexx/app/_server.py
@@ -11,7 +11,7 @@ from .. import config
 _current_server = None
 
 
-def create_server(host=None, port=None, new_loop=False, backend='tornado'):
+def create_server(host=None, port=None, new_loop=False, backend='tornado', **server_kwargs):
     """
     Create a new server object. This is automatically called; users generally
     don't need this, unless they want to explicitly specify host/port,
@@ -33,6 +33,7 @@ def create_server(host=None, port=None, new_loop=False, backend='tornado'):
             which is made current when ``start()`` is called. If ``False``
             (default) will use the current IOLoop for this thread.
         backend (str): Stub argument; only Tornado is currently supported.
+        **server_kwargs: keyword arguments passed to the server constructor
     
     Returns:
         server: The server object, see ``current_server()``.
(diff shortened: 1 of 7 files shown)


almarklein:
Thanks!

I think it would be nice to document this feature. I think the best place might be to add a few lines to the docstring of `create_server()` on how `server_kwargs` can be used to enable SSL.

Is it also possible to add an example script to demo this? Not sure what cert and keyfile to point to though. What do you think? 

I see a few more places that use `http`:
* The `url` prop in `app/_app.py`
* `init_interactive()` in `app/_funcs.py`

almarklein:
Travis fails because of style errors. You can run the style test locally by running `python make test style` in the root flexx directory.

Korijn:
Should the port be set to `443` automatically if `None` for the `https` protocol? I'm not sure if it will work on other ports.

pigay:
@Korijn from my tests, it works fine for default port (49190)

alexveden:
@pigay thx for your work. Have you added new config setting handling for HTTPS connections?

pigay:
This error (https://travis-ci.org/zoofIO/flexx/jobs/181720106) seems to happen in python2.

I don't know much about python testing. How can I debug this error?

pigay:
@alexveden not yet. I was thinking of it, though.

However, this configuration is specific to Tornado. Should I add `tornado_certfile` and `tornado_keyfile` in `flexx._config.py`?

alexveden:
@pigay It would be great, because the problem is also related to remote Ipython notebooks too https://github.com/zoofIO/flexx/issues/299

alexveden:
@pigay Am I correct, do we need both key and cert to connect SSL, but according your last commit https://github.com/zoofIO/flexx/pull/307/commits/ccfa4b1dfa79e6ba355fbd540e1b84c4237a5f71 its allowed to add one of those files.

pigay:
@alexveden I think so. I found easier to handle each file independently and error messages from tornado should still be meaningful.

pigay:
A question: I'm not really fluent with Github. Do I need to create another PR after rebase to solve the merge conflict?

almarklein:
@pigay you can create a new branch if you prefer, but it it nor necessary. On a merge conflict in a PR you can either merge your branch with master, or rebase it on master. Preference depends on the project, I slightly prefer the latter:
```
git checkout master
git pull
git checkout your-branch
git rebase master
... resolve conflicts
git add ... files that were in conflict
git rebase --continue
git push your-remote your-branch -f
```
The `-f` force-pushes, i.e. overwrites history. Never ever use that on a `master` branch :P

pigay:
@almarklein the `-f` what was I was missing. Thanks.

alexveden:
@almarklein how long could it take to merge this pull request to the master? Thx

flexx/app/_tornadoserver.py
```diff
@@ -63,17 +63,30 @@ def _get_io_loop(self):
...
                 self._loop = IOLoop(make_current=True)
     
-    def _open(self, host, port):
+    def _open(self, host, port, **kwargs):
         # Note: does not get called if host is False. That way we can 
         # run Flexx in e.g. JLab's application.
-        
+
+        # handle ssl, wether from configuration or given args
+        if 'ssl_certfile' in config and config.ssl_certfile:
```
almarklein: That first test `ssl_certfile' in config` should not be necessary

flexx/app/_server.py
```diff
@@ -33,9 +34,24 @@ def create_server(host=None, port=None, new_loop=False, backend='tornado'):
...
+    
+        .. code-block:: py
+        
+            app.create_server(ssl_options = {'certfile' : '/path/to/certfile',
+                                             'keyfile' : '/path/to/keyfile'})
+            app.serve(Example, 'Example')
+            app.run()
+
+        Alternately, cert and key files can be provided through
+           ``ssl_certfile`` and ``ssl_keyfile`` configuration variables.
```
almarklein: I like the idea of setting ssl options via the config. In fact, I think we should advocate that as *the* way to do it. We can keep the `server_kwargs`, but can you please please remove the example in this docstring (I know I asked you to add it earlier, sorry!).

flexx/app/examples/serve_ssl.py
```diff
@@ -0,0 +1,33 @@
...
+
+# Some very secret Model
+class Example(ui.Widget):
+    def init(self):
+        ui.Button(text='Secret Button')
+
+# configure web server for SSL
+app.create_server(ssl_options={'certfile' : CERTFILE,
+                               'keyfile' : KEYFILE})
+
```
almarklein: Can you please use `flexx.config` here instead? I think that should be the approach to put forward as the way to enable https.

flexx/app/examples/serve_ssl.py
```diff
@@ -0,0 +1,33 @@
...
+from flexx import app, ui
+
+# generate self-signed certificate for this example
+import os
+
+CERTFILE = '/tmp/self-signed.crt'
+KEYFILE = '/tmp/self-signed.key'
+
+os.system('openssl req -x509 -nodes -days 1 -batch -newkey rsa:2048 '
+          '-keyout %s -out %s' % (KEYFILE, CERTFILE))
```
almarklein: Out of curiosity, if Flexx ever supports a web framework alternative to Tornado, would the same key and cert files be usable?

almarklein:
Thanks @pigay !

pigay:
My pleasure!

PR #309, opened 2016-12-06
Fixed string replace in events.

Maxyme:
pscripted code read:

```
else if (_pyfunc_op_equals(selector, "*")) {
            t = "Invalid connection \"%s\" because %s is not a tuple/list.";
            err_3 = new Error('RuntimeError:' + (t % [obname_full, obname])); err_3.name = "RuntimeError"; throw err_3;
```

Which raised a NaN error on Chrome when trying to print this:
_"Invalid connection volumes* because volumes is not a tuple/list."_
It was instead doing this (in the js console):
```
t = "Invalid connection \"%s\" because %s is not a tuple/list."; # "Invalid connection "%s" because %s is not a tuple/list."
t % [obname_full, obname]  # returns NaN
```
This should fix this error.

### DIFF
--- a/flexx/event/_handler.py
+++ b/flexx/event/_handler.py
@@ -384,7 +384,9 @@ class Handler:
                 self._seek_event_object(index, path, sub_ob)
             return
         elif selector == '*':  # "**" is recursive, so allow more
-            t = "Invalid connection {obname_full} because {obname} is not a tuple/list.";
-            raise RuntimeError(t.replace("{opname_full}", obname_full).replace("{opname}", obname))
+            t = "Invalid connection {obname_full} because {obname} \
+                is not a tuple/list."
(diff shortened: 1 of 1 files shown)


ivoflipse:
If only flexx supported formatting strings using `.format()`...

almarklein:
> If only flexx supported formatting strings using .format()...

That should be doable, by implementing a JS variant of `format()`. The current formatter is implemented by the transpiler. Maybe even better would be to support `f"haha {some_var}"`.

almarklein:
> The current formatter is implemented by the transpiler. 

Arg, yes, and that's why the error was there, because the transpiler is not smart enough to know what `t` is.

almarklein:
Thanks!

PR #313, opened 2016-12-08
Support for sending data to JS + improved security

### LINKED ISSUE #143: Provide means to send large (binary) data between Python and JS
One intention of Flexx is to allow writing scientific-oriented apps, to make it easy to share ideas and results. Science needs data, e.g. nd arrays.

Possibilities:
- Send the data over the existing websocket. Downsides: we probably want to have a binary websocket then, and this implies that in exported apps, the commands are base64 encoded inside the html :/
- Send the data over an auxiliary websocket
- Send the data via Ajax. Seems to work great in Arbiter. This can be done in binary form and provides a similar mechanic to load data from the web instead of from Python. Can this made to work for exported apps (i.e. can the blob be loaded from a static file next to the HTML doc)?

### LINKED ISSUE #297: video widget local file
Is there a way to get the video widget to play mp4 files located locally? I've tried every format for the url I could think of but the video isn't playing. I'm not too familiar with browser tech so I'm probably trying to break some rule.

Ultimately I'd like to be able to dump mp4 files into a folder locally and present a list of files in that folder to the user and when the user clicked on a video it would play in a video widget. Any help on getting me moving would be greatly appreciated.

almarklein:
Fixes #143, fixes #297

* Adds `Model.send_data(data, meta)` with data being `bytes` or a URI, which invokes ...
* Adds `Model.JS.retrieve_data(url, meta)` which retrieves binary data using AJAX. When data arrives, it invokes ...
* Adds `Model.JS.receive_data(data, meta)` which should be overloaded by classes that want to process binary data.
* All Ajax for now. later on we can use a dedicated binary websocket when `data` in `send_data()` is not a URI.
* All about bytes for now. Later on we can see if we want to support higher level data structures like arrays.
* ~~Also got a static file server that provides access to data URI's staring with 'file://'.~~
* ~~Using data (via `Model.send_data`, `Session.add_data` or `app.assets.add_shared_data`) that is local to the filesystem (i.e. 'file://') is only allowed for files inside pre-defined "data dirs".~~
* Remove functionality to load assets or data via local files; users need to `open()` the data themselves.
* Add example to setup Tornado static file server.


flexx/app/_assetstore.py
```diff
@@ -433,16 +446,49 @@ def add_shared_data(self, name, data):
...
-                data = open(data.split('//', 1)[1], 'rb').read()
-            elif data.startswith(('http://', 'https://')):
-                data = urlopen(data, timeout=5.0).read()
-        if not isinstance(data, bytes):
-            raise TypeError('add_shared_data() data must be bytes.')
+        if isinstance(data, str) and data.startswith('file://'):
+            if not self.in_data_dir(data):
+                raise ValueError('add_shared_data() with a local file must exist '
+                                 'and be inside a data dir (see add_data_dir().')
+        elif isinstance(data, str) and data.startswith(('http://', 'https://')):
```
almarklein: @korijn I was wondering if you have suggestions w.r.t. security here. There actually was a vulnerability where the `ImageWidget` uses `add_data()` when its source is a local filename, which can be exploited to watch *any* file on the server.

Korijn: Make sure static file server can only serve files from a specific root path!

Korijn: Make sure static file server can only serve files from a specific root path!

almarklein: That's what this does, except I called them data dirs and there can be multiple. I think we're good.

Korijn: https://docs.djangoproject.com/en/1.10/howto/static-files/

> This method is grossly inefficient and probably insecure, so it is unsuitable for production.

Korijn:
Perhaps a silly comment, but I think the file:// protocol does not work anywhere else besides locally on the client's machine? I mean, it doesn't specify a host or port right?

almarklein:
Data by pointing to a file using `file://` are files on the *server*. When you're running Flexx locally, its the same file system, of course.

* For `session.add_data()` and `app.assets.add_shared_data()` it provides a means to select a file (on the server) for the client to access via the url that these functions return.
* For `Model.send_data()` it provides a means to send thet contents of a file on the server to the client.

So the client will never use a url starting with `file://`. Exported apps use a relative path to the data, so that when an exported app is served by a file server (e.g the docs!) it actually still works. That's pretty neat; you can create an app with shows scientific data interactively, export it, and put it on your blog put it on S3 etc.

I will try to improve the docs and comments to make this a bit more clear.

flexx/app/_assetstore.py
almarklein: I understand, and agree that we should inform our users accordingly, but being able to serve static files from the server is common enough that we should support it, I think. E.g. if I want to put some images in my (non-production) app, I would not want to have to read them in memory, or setup a tornado StaticFileHandler (let along nginx).

To summarize the current approach:
- users can specify "data dirs" using `app.assets.add_data_dir()`. Maybe this needs to be `flexx.config.data_dirs`?
- When a `file://....` is specified in `send_data`, `add_data` or `add_shared_data`, it is verified that the given file exists and is inside a data dir.
- When the file is requested (via `/flexx/static/...`), and handled by the StaticFileHandler, the real filename on the FS is looked up in the assetstore or session, and it is again verified that the file is inside a data dir.

almarklein:
Summary of chat with @Korijn: always adding a static file server to Flexx is a potential risk, even if we get our own code airtight, because we cannot guarantee that the server itself is 100% safe, which would make Flexx unsuited for production. A static file server should be opt-in.

So I am going to:

* remove the static file server again.
* add a util, or at least an example, to use Tornado's `StaticFileServer`.
* remove any convenience related to loading local files using "file://".
* scan the codebase for `isfile` `file://` and `open(` to make sure there are no more places for this kind of vulnerability.

It will mean that users can can still use data on their (server's) filesystem, but they'd have to either:
    *  `open(..).read()` the data. (exported apps will work as long as they're "served", e.g. on s3 or readthedocs)
    * put the data in a place where its accessible with `http` (e.g. dropbox). (exported apps Just Work)
    * setup a static file server, e.g. a Tornado `StaticFileServer`. (cannot export such an app)

PR #314, opened 2016-12-09
added disabled property to button

Maxyme:
Added a property to set buttons to disabled and reverse.

### DIFF
--- a/flexx/ui/widgets/_button.py
+++ b/flexx/ui/widgets/_button.py
@@ -3,18 +3,18 @@
 Simple example:
 
 .. UIExample:: 50
-    
+
     b = ui.Button(text="Push me")
 
 
 Example with interaction:
 
 .. UIExample:: 200
-    
+
     from flexx import app, ui, event
-    
+
     class Example(ui.BoxPanel):
-        
+
         def init(self):
             with ui.VBox():
                 self.b1 = ui.Button(text='apple')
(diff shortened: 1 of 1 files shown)


flexx/ui/widgets/_button.py
```diff
@@ -115,11 +121,18 @@ def _init_phosphor_and_node(self):
...
-        
+
         @event.connect('text')
         def __text_changed(self, *events):
             self.node.innerHTML = events[-1].new_value
 
+        @event.connect('disabled')
+        def __disabled_changed(self, *events):
+            if events[-1].new_value:
+                self.node.setAttribute("disabled", "disabled")
```
ivoflipse: Setting `"disabled"` to `"disabled"` wouldn't that be a bool or does the value not matter anyway?

ivoflipse:
```
<input disabled />
<input disabled="" />
<input disabled="disabled" />
<input disabled="true" />
<input disabled="false" /> <!-- still disabled! -->
```
Guess the value indeed doesn't matter

Maxyme: it's a funny property:
https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

almarklein:
Thanks!

flexx/ui/widgets/_button.py
Korijn: It's just how HTML works :)

PR #317, opened 2016-12-11
SSL support for remotely running Jupyter notebooks

alexveden:
* app.init_notebook() fix added WSS to url
* TornadoServer  Comparing hostnames in case of Jupyter notebooks connecting_host and serving_host have same host names but different ports

### DIFF
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -9,5 +9,6 @@ config = Config('flexx', '~appdata/.flexx.cfg',
                          'it is closed.'),
     ssl_certfile=('', str, 'cert file for https server.'),
     ssl_keyfile=('', str, 'key file for https server.'),
-    hosts_whitelist=('', str, 'Comma separated list of allowed <host>:<port> values to pass cross-origin checks.'),
+    hosts_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
+                              'values to pass cross-origin checks.'),
     )
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -1,14 +1,14 @@
 from .util.config import Config
 
 config = Config('flexx', '~appdata/.flexx.cfg',
-    log_level=('info', str, 'The log level to use (DEBUG, INFO, WARNING, ERROR)'),
-    hostname=('localhost', str, 'The default hostname to serve apps.'),
-    port=(0, int, 'The default port to serve apps. Zero means auto-select.'),
-    webruntime=('', str, 'The default web runtime to use. Default is xul/browser.'),
-    ws_timeout=(20, int, 'If the websocket is idle for this amount of seconds, '
+                log_level=('info', str, 'The log level to use (DEBUG, INFO, WARNING, ERROR)'),
+                hostname=('localhost', str, 'The default hostname to serve apps.'),
+                port=(0, int, 'The default port to serve apps. Zero means auto-select.'),
+                webruntime=('', str, 'The default web runtime to use. Default is xul/browser.'),
+                ws_timeout=(20, int, 'If the websocket is idle for this amount of seconds, '
                          'it is closed.'),
-    ssl_certfile=('', str, 'cert file for https server.'),
-    ssl_keyfile=('', str, 'key file for https server.'),
-    hosts_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
+                ssl_certfile=('', str, 'cert file for https server.'),
+                ssl_keyfile=('', str, 'key file for https server.'),
+                host_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
                               'values to pass cross-origin checks.'),
-    )
+                )
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -9,4 +9,5 @@ config = Config('flexx', '~appdata/.flexx.cfg',
                          'it is closed.'),
     ssl_certfile=('', str, 'cert file for https server.'),
     ssl_keyfile=('', str, 'key file for https server.'),
+    hosts_whitelist=('', str, 'Comma separated list of allowed <host>:<port> values to pass cross-origin checks.'),
     )
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -1,14 +1,14 @@
 from .util.config import Config
 
 config = Config('flexx', '~appdata/.flexx.cfg',
-                log_level=('info', str, 'The log level to use (DEBUG, INFO, WARNING, ERROR)'),
-                hostname=('localhost', str, 'The default hostname to serve apps.'),
-                port=(0, int, 'The default port to serve apps. Zero means auto-select.'),
-                webruntime=('', str, 'The default web runtime to use. Default is xul/browser.'),
-                ws_timeout=(20, int, 'If the websocket is idle for this amount of seconds, '
-                         'it is closed.'),
-                ssl_certfile=('', str, 'cert file for https server.'),
-                ssl_keyfile=('', str, 'key file for https server.'),
-                host_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
-                              'values to pass cross-origin checks.'),
-                )
+            log_level=('info', str, 'The log level to use (DEBUG, INFO, WARNING, ERROR)'),
+            hostname=('localhost', str, 'The default hostname to serve apps.'),
+            port=(0, int, 'The default port to serve apps. Zero means auto-select.'),
+            webruntime=('', str, 'The default web runtime to use. Default is xul/browser.'),
+            ws_timeout=(20, int, 'If the websocket is idle for this amount of seconds, '
+                     'it is closed.'),
+            ssl_certfile=('', str, 'cert file for https server.'),
+            ssl_keyfile=('', str, 'key file for https server.'),
+            host_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
+                          'values to pass cross-origin checks.'),
+            )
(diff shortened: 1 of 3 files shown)


flexx/app/_funcs.py
```diff
@@ -189,7 +189,12 @@ def init_notebook():
     
     # We set session_id and app_name in a way that it does not end up
     # in exported notebook.
-    url = 'ws://%s:%i/flexx/ws/%s' % (host, port, session.app_name)
+    if config.ssl_certfile != '' or config.ssl_keyfile != '':
+        # Connect over SSL
+        url = 'wss://%s:%i/flexx/ws/%s' % (host, port, session.app_name)
+    else:
+        url = 'ws://%s:%i/flexx/ws/%s' % (host, port, session.app_name)
```
almarklein: Better use the same method as elsewhere, using `server.proto`:
```py
url = '%s://%s:%i/flexx/ws/%s' % (server.proto, host, port, session.app_name)
```

flexx/app/_tornadoserver.py
```diff
@@ -678,13 +678,18 @@ def check_origin(self, origin):
...
         if serving_hostname == 'localhost':
             return True  # Safe
         elif serving_hostname == '0.0.0.0':
             return True  # we cannot know if the origin matches
         elif serving_host == connecting_host:
             return True
+        elif serving_hostname == connecting_hostname:
+            # Comparing hostnames in case of Jupyter notebooks
+            # connecting_host and serving_host have same host names but different ports
+            return True
```
almarklein: Is this something that happens with https?

alexveden: Tornado refuses connection on same host but with different ports, this is common case with remote notebook.

almarklein: So, this is not specific to httpS?

flexx/app/_funcs.py
alexveden: > Better use the same method as elsewhere, using server.proto
I guess the proto property returns 'http' / 'https', but we need 'ws' and 'wss'. Am I wrong?

Korijn: Typically the websocket en http protocols are configured independently. A `server.ws_proto` attribute might make sense?

almarklein: What would be a use-case for using a mix of `http` and `https`? Would it not make sense to just use ssl everywhere if a certificate is available?

flexx/app/_tornadoserver.py
Korijn: CORS :) any connections established by JS (ajax, websocket) must be on the same Origin (protocol+hostname+port), unless the remote server explicitly permits your Origin to communicate with them.

alexveden: > The error should actually indicate this in the browser console: is that indeed the case @alexveden ?
Correct

Korijn: There appears to be some sort of whitelisting mechanism: https://github.com/jupyter/notebook/pull/493

alexveden: Maybe the better solution will be in comparison between config.hostname and connecting_hostname?

flexx/app/_funcs.py
almarklein: @alexveden ah, you're right. In that case: `proto = 'wss' if server.proto == 'https' else 'ws'`

flexx/app/_tornadoserver.py
alexveden: > Taking one step back, why is the port number different? I have not experienced this before, can we do something to use the same port?

Typically Jupyter notebook is running on 8888 port, and flexx backend is on 49xxx port, when you are trying to connect to WebSocket from running notebook the ports are different, so this lead to CORS checks failure.

almarklein: Ah, right, and I never get the error because I always try on localhost :)  This should be fixed. Looking into it.

almarklein: ah ... that wont work, of course. The browser is connected to the Jupyter server via `host:8888`. The server and kernel maintain a ZMQ connection. Flexx needs its own connection (between browser and kernel), and thus its own port. 

Did you use `config.hostname` to make Flexx use the same hostname as the Jupyter server? I guess that would be the way to do it, since AFAIK Flexx does not have a way (from inside the kernel) to know what hostname the server is using.

alexveden: >Did you use config.hostname to make Flexx use the same hostname as the Jupyter server?

Correct, I'm using this option to make it work on the remote server.

almarklein: What about this:
```py
    def check_origin(self, origin):
        """ Handle cross-domain access; override default same origin policy.
        """
        # http://www.tornadoweb.org/en/stable/_modules/tornado/websocket.html
        #WebSocketHandler.check_origin
        
        serving_host = self.request.headers.get("Host")
        serving_hostname, serving_port = serving_host.split(':', 1)
        connecting_host = urlparse(origin).netloc
        connecting_hostname, connecting_port = connecting_host.split(':', 1)
        
        # Whitelist of ports - only if interactive mode is on, e.g. notebook
        ok_ports = []
        if manager.has_app_name('__default__'):
            ok_ports = '80', '8888'
        
        if serving_hostname == 'localhost':
            return True  # Safe
        elif serving_host == connecting_host:
            return True  # Passed most strict test, hooray!
        elif serving_hostname == '0.0.0.0' and serving_port == connecting_port:
            return True  # We cannot know if the hostname matches
        elif serving_hostname == connecting_hostname and connecting_port in ok_ports:
            return True  # Hostnames match, and we find these ports ok
        else:
            logger.warn('Connection refused from %s' % origin)
            return False
```

alexveden: `if manager.has_app_name('__default__'):
            ok_ports = '80', '8888'`
this is ok, but ok_ports should be configurable from file, because if there are several instances of Jupyter server running they are serving different ports, for example 8889, 8890 and etc

almarklein: Make that `flexx.config.cors_whitelist`, which is a list of `"hostname:port"` strings. That way we also fix it for when the notebook server is on another machine ... "future ready" :)

thinking about enforcing the correct protocol too ... will update shortly

flexx/app/_tornadoserver.py
```diff
@@ -676,19 +676,17 @@ def check_origin(self, origin):
         #WebSocketHandler.check_origin
         
         serving_host = self.request.headers.get("Host")
-        serving_hostname = serving_host.split(':')[0]
-        connecting_host = urlparse(origin).netloc
-        connecting_hostname = connecting_host.split(':')[0]
+        serving_hostname = serving_host.split(':')[0].lower()
+        connecting_host = urlparse(origin).netloc.lower()
+
```
alexveden: I'm not sure about .lower() the host names, but I think it could be useful in some cases.

flexx/app/_tornadoserver.py
```diff
@@ -676,15 +676,18 @@ def check_origin(self, origin):
...
-        serving_hostname = serving_host.split(':')[0]
-        connecting_host = urlparse(origin).netloc
+        serving_hostname = serving_host.split(':')[0].lower()
+        connecting_host = urlparse(origin).netloc.lower()
+
         
         if serving_hostname == 'localhost':
             return True  # Safe
         elif serving_hostname == '0.0.0.0':
             return True  # we cannot know if the origin matches
```
almarklein: While you're in here, can you please add the test for equal ports for the `0.0.0.0` case as I wrote in the example?
I'll see what to do with the protocol at a later time.

flexx/app/_tornadoserver.py
```diff
@@ -676,15 +676,18 @@ def check_origin(self, origin):
...
         elif serving_host == connecting_host:
             return True
+        elif connecting_host in config.hosts_whitelist.lower():
```
almarklein: `connecting_host.lower()`? Or just make it case sensitive?

flexx/app/_tornadoserver.py
alexveden: Sure where is the test file located? Or I need to create new one?

flexx/app/_tornadoserver.py
almarklein: http://stackoverflow.com/questions/7666782/are-idn-domain-names-case-sensitive

My feeling is to leave the case. It might matter in some cases, and the "convenience" that we provide by making it insensitive is minimal.

flexx/_config.py
```diff
@@ -9,4 +9,6 @@
                          'it is closed.'),
     ssl_certfile=('', str, 'cert file for https server.'),
     ssl_keyfile=('', str, 'key file for https server.'),
+    hosts_whitelist=('', str, 'Comma separated list of allowed <host>:<port> '
+                              'values to pass cross-origin checks.'),
```
almarklein: Nitpicking, but I'd prefer `host_whitelist`, because "list" already implied plural, and googling it seems to suggests a few projects also using that name.

flexx/app/_tornadoserver.py
alexveden: Sorry, I'm not getting the point...

> elif serving_hostname == '0.0.0.0':

When serving_hostname can be 0.0.0.0?

almarklein: A socket (i.e. server) can bind to `0.0.0.0` which means that any connection is accepted that is targeting the machine (e.g. 'localhost' works, but also the IP for the LAN connection and the IP for the WLAN connection).

Anyway, I'll merge this now. I'm planning a few tweaks areound `flexx.app` anyway.

almarklein:
Thanks @alexveden !

PR #318, opened 2016-12-12
Tweaks to play nice with jlab

almarklein:


alexveden:
Just wondering, are these changes compatible with vanilla Jupyter notebooks?

almarklein:
Yes, this should not get into the way of running "regular" notebooks. These are to make the flexx embeddable in Jlab as a tool to make plugins. A next step would be be to allow running Flexx inside notebooks in jlab, which would require multiple Flexx "sessions" on the same page.

PR #322, opened 2016-12-14
Add assetstore.remove_data()

almarklein:
A possible risk of adding this function is that it becomes easier to abuse the asset store as an in-memory static file server of sorts. Is that bad? Probably not for small-scale things, but in production its likely a bad idea.

### DIFF
--- a/flexx/app/_assetstore.py
+++ b/flexx/app/_assetstore.py
@@ -415,8 +415,12 @@ class AssetStore:
     def add_shared_data(self, name, data):
         """ Add data to serve to the client (e.g. images), which is shared
         between sessions. It is an error to add data with a name that is
-        already registered. See ``Session.add_data()`` to set data per-session
-        and `Model.send_data()`` to send data to Model objects directly.
+        already registered. The data remains available until it is removed via
+        ``remove_data()`` or until the server stops.
+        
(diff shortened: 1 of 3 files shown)


JohnLunzer:
Unless the performance is _terrible_ then "abuse" is not something to worry about. It's possible that this pattern is the least complex method for temporarily bringing in assets. That isn't bad, but a feature allowing fast prototyping.

It would be a good idea to make a note in the docs that there are performance implications when using the asset store as a pseudo static file server. This can be seen when using the asset store without a static file server to load video for playing. Not only must the file be entirely loaded into memory but it also prevented the video from streaming using the HTML5 video element. 

With my HTML5 video example it was important for me to see things working using the less complex method because it helped me quickly get up and running. Once I saw what was possible I worked towards setting up the Tornado static file server to get the performance I was looking for.

In most cases the optimum performance isn't necessary, I think this feature reflects that.

almarklein:
Well, for get something up quickly, `app.assets.add_data()` already works. With a `remove_data()` method, it becomes possible to "manage" the data better, but perhaps the need for such a method should be the point where the user should switch to a more scalable solution :)

ivoflipse:
@almarklein do you plan on doing anything with this PR?

almarklein:
We should probably just close it. We can always re-open if needed.

PR #323, opened 2016-12-15
Add example for code editor widget based on codemirror

almarklein:
This could later be extended (e.g. get properties to toggle line numbers, syntax highlighting, etc.) and put as a widget in the library.

### DIFF
--- a/flexx/ui/examples/code_editor.py
+++ b/flexx/ui/examples/code_editor.py
@@ -1,0 +1,53 @@
+# doc-export: CodeEditor
+"""
+This example demonstrates a code editor widget based on CodeMirror.
+"""
+
+# todo: Maybe this should be a widget in the library (flexx.ui.CodeMirror) ?
+
+from flexx import ui, app, event
+from flexx.pyscript.stubs import window
+
+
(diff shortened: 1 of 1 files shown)


JohnLunzer:
This is great thanks. I can run the example and see the line number gutters, but I'm not getting any code highlighting. I tested changing the mode string to another language but still unhighlighted.

almarklein:
The `mode` properety is probably incorrect:

The mode to use. When not given, this will default to the first mode that was loaded. It may be a string, which either simply names the mode or is a MIME type associated with the mode. Alternatively, it may be an object containing configuration options for the mode, with a name property that names the mode (for example {name: "javascript", json: true}). The demo pages for each mode contain information about what configuration parameters the mode supports. You can ask CodeMirror which modes and MIME types have been defined by inspecting the CodeMirror.modes and CodeMirror.mimeModes objects. The first maps mode names to their constructors, and the second maps MIME types to mode specs.

JohnLunzer:
Mode property is fine. Must add 
```python
app.assets.associate_asset(__name__, base_url + '5.21.0/mode/python/python.js') 
```
To the example. If you look at base codemirror.js it knows nothing about python.

PR #325, opened 2016-12-15
add disabled property to LineEdit

pigay:
Copying #314 code to LineEdit widget.

This could probably apply to all widgets based on a `<input>` tag.

Can we just do the same for BaseDropDown widget?

### DIFF
--- a/flexx/ui/widgets/_color.py
+++ b/flexx/ui/widgets/_color.py
@@ -40,7 +40,13 @@ class ColorSelectWidget(Widget):
                 raise ValueError('%s.color must be in #rrggbb format, not %r' %
                                  (self.id, v))
             return str(v)
-    
+
+        @event.prop
+        def disabled(self, v=False):
+            """ Whether the color select is disabled.
+            """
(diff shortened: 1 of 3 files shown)


almarklein:
> This could probably apply to all widgets based on a <input> tag.

That leaves the `Slider` and the `ColorSelectWidget`. Do you want to look into these in this PR?

The DropDown is a custom widget, so that would need a little more work.

Korijn:
Isnt the dropdown a simple wrapper around <select>?

almarklein:
No, the combobox has a text `<input>`, but the dropdown part is all custom.

pigay:
Don't know how to do it for dropdown...

almarklein:
> Don't know how to do it for dropdown...

That's ok. I'd need to sit down for that and fiddle with some CSS, I think. Let's go with what we have for now.

Is this ready from your end?

pigay:
Yes, nothing to add from my part.

almarklein:
Thanks!

PR #326, opened 2016-12-15
make CodeMirror example more impressive

JohnLunzer:
This PR fixes the Python syntax not being highlighted.

It also makes the example a little more impressive by showing how easy it is to change themes and enable features.

### DIFF
--- a/flexx/pyscript/parser3.py
+++ b/flexx/pyscript/parser3.py
@@ -172,10 +172,10 @@ Using JS specific functionality
 When writing PyScript inside Python modules, we recommend that where
 specific JavaScript functionality is used, that the references are
 prefixed with ``window.`` Where ``window`` represents the global JS 
-namespace to prefix JS objects. ``window`` will simply be removed
-by the transpiler. This helps make it clear that the functionality 
-is specific to JS, and also helps static code analysis tools like 
-flake8.
+namespace. All global JavaScript objects, functions, and variables
+automatically become members of the ``window`` object. This helps
+make it clear that the functionality is specific to JS, and also
+helps static code analysis tools like flake8.
 
 .. pyscript_example::
     
--- a/flexx/pyscript/parser3.py
+++ b/flexx/pyscript/parser3.py
@@ -166,14 +166,16 @@ Str methods
     "foobar".upper()
 
 
-Using JS specicic functionality
+Using JS specific functionality
 -------------------------------
 
 When writing PyScript inside Python modules, we recommend that where
 specific JavaScript functionality is used, that the references are
-prefixed with ``window.``. This helps make it clear that the
-functionality is specific to JS, and also helps static code analysis
-tools like flake8.
+prefixed with ``window.`` Where ``window`` represents the global JS 
+namespace to prefix JS objects. ``window`` will simply be removed
+by the transpiler. This helps make it clear that the functionality 
+is specific to JS, and also helps static code analysis tools like 
+flake8.
 
 .. pyscript_example::
     
--- a/flexx/ui/examples/code_editor_ace.py
+++ b/flexx/ui/examples/code_editor_ace.py
@@ -1,0 +1,44 @@
+# doc-export: CodeEditor
+"""
+This example demonstrates a code editor widget based on Ace.
+"""
+
+# todo: Maybe this should be a widget in the library (flexx.ui.Ace) ?
+
+from flexx import ui, app, event
+from flexx.pyscript import window
+
+# Associate Ace's assets with this module so that Flexx will load
+# them when (things from) this module is used.
+base_url = 'https://cdnjs.cloudflare.com/ajax/libs/ace/1.2.6/'
+app.assets.associate_asset(__name__, base_url + 'ace.js')
+app.assets.associate_asset(__name__, base_url + 'mode-python.js')
+app.assets.associate_asset(__name__, base_url + 'theme-solarized_dark.js')
+
+class CodeEditor(ui.Widget):
+    """ A CodeEditor widget based on Ace.
+    """
+    
+    CSS = """
+    .flx-CodeEditor > .ace {
+        width: 100%;
+        height: 100%;
+    }
+    """
+    
+    class JS:
+        def init(self):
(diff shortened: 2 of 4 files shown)


JohnLunzer:
Also hoping by fixing this and showing how nice it can look that it will help convince @almarklein to make this an "official" widget.

JohnLunzer:
I'm impressed with how this small class definition creates a fully functional and feature rich editor widget.

CodeMirror's editor API is almost entirely accessible through `self.cm` in this example. Some niceties could certainly be added to the Python API for a CodeMirrorWidget but right now the `class JS` side should get people going.

almarklein:
Thanks John, I like how you took this to the next level!

I agree that a widget like this is generally useful. However, there are a few things holding me back to include this widget in the library:

* Multiple assets are needed, and it depends on the use-case which ones (and including all would probably make it a bit heavy).
* It would also be nice if the assets were modules to avoid dumping stuff on `window`. Maybe CodeMirror does support this and we just need different assets? Or we can pre-package something ourselves similar to `phosphor-all`.
* As it is now, any GUI that makes use of this widget would need an internet connection to download the assets. Users might not expect this, so we probably want a local copy. We can certainly do this (we do the same for PhosphorJS) but ...
* Is CodeMirror the most suitable library that we can use? I don't have a clear view on this.

Considering that it quite easy to set up, I think that for the time being this example should be enough to get people going. It could still be extended a bit by adding some properties to dynamically set certain options, or make CodeMirror events available as Flexx events, but we should probably not do too much of that in order to keep the example simple.

JohnLunzer:
@almarklein, I agree with your assessment. Like you said I think we would have to pull in a CodeMirror-all to make this viable for non-internet users. As it is still useful.

Also. I'm not sure why it's failing the check after my second minor change.

almarklein:
The fail is a glitch. The `test_live` test occasionally fails for no apparent reason. I just restarted it.

JohnLunzer:
@almarklein, it looks like [ace](https://ace.c9.io/#nav=about) is another heavy hitter. I'll probably take a look at that as an option as well. I'm sure there are a plethora of others but Ace and CodeMirror are the two top hits when googling "javascript code editor".

JohnLunzer:
Added an example with Ace. 

I think we could settle on either Ace _or_ CodeMirror as an "official" widget. They have similar feature sets. They have similar organization, as in lots of individual assets. They are at a similar level of maturity and based on their git activity they are in a similar stage of their software lifecycle. CodeMirror's API seems slightly more complex than Ace; Ace does more for you automatically.

For a few reasons, I'm _initially_ inclined to tilt more towards CodeMirror:

- I like the use of css files for styles instead of how Ace wraps the CSS in a js with some automation. From the point of view of designing an IDE it feels cleaner to have the CSS file for styling the rest of the API to match the code editor. The Ace API for loading styles could potentially change and more easily break other parts of the IDE.
- CodeMirror is still slightly more active.
- CodeMirror has a Cython syntax highlighting mode. Although I want flexx and potential IDEs built with it to be as generic as possible, if we're being honest we're more likely to attract Python developers and a Cython syntax highlighter out of the box is nice.

These are just a few superficial reasons. I've by no means dug into either library in any detail. I think it's okay to take time in deciding what to choose a library to potentially pull into Flexx.

flexx/pyscript/parser3.py
```diff
@@ -166,14 +166,16 @@
...
 When writing PyScript inside Python modules, we recommend that where
 specific JavaScript functionality is used, that the references are
-prefixed with ``window.``. This helps make it clear that the
-functionality is specific to JS, and also helps static code analysis
-tools like flake8.
+prefixed with ``window.`` Where ``window`` represents the global JS 
+namespace to prefix JS objects. ``window`` will simply be removed
+by the transpiler. This helps make it clear that the functionality 
+is specific to JS, and also helps static code analysis tools like 
+flake8.
```
JohnLunzer: Sorry to slip this into the PR as it's slightly unrelated.

JohnLunzer:
The more I ponder this, the more I believe that it's not in Flexx's scope to make these types of widgets "official". CodeMirror and Ace have extensive APIs with myriad options and configurations. Flexx could never hope to faithfully expose these APIs satisfactorily on the Python side. This goes back to something @Korijn said about how even trying to keep up with the API for the HTML5 video widget could become cumbersome. My feeling is that Flexx's scope is to make wrapping libraries like CodeMirror and Ace as easy and as intuitive as possible. 

I'm already pretty impressed with how easy and intuitive it's been. Continuing to work with these libraries within Flexx will expose where improvements can be made. If someday there is popular demand for pulling in a library dependency for an official CodeEditorWidget we could revisit. Until then I'm pretty satisfied with these examples and look forward to integrating them into an IDE design.

almarklein:
>  Flexx could never hope to faithfully expose these APIs satisfactorily on the Python side. [...] Flexx's scope is to make wrapping libraries like CodeMirror and Ace as easy and as intuitive as possible.

Well said.

Note that the editors actually *are* importable using e.g. `from flexx.ui.examples.code_editor_ace import CodeEditor_Ace`. Though if someone plans to build anything serious with them, its probably better to just copy the code.

flexx/ui/examples/code_editor_ace.py
```diff
@@ -0,0 +1,44 @@
...
+    class JS:
+        
+        def init(self):
+            # https://ace.c9.io/#nav=api
+            self.ace = window.ace.edit(self.node, "editor")
+            self.ace.setValue("import os\n\ndirs = os.walk")
+            self.ace.navigateFileEnd()  # otherwise all lines highlighted
+            self.ace.setTheme("ace/theme/solarized_dark")
+            self.ace.getSession().setMode("ace/mode/python")
+        @event.connect('size')
```
almarklein: While you're making that change, an newline between the two methods here would be nice :)

JohnLunzer:
I think everything should be in order now. 

Ignore the commit message on the last commit, I just forgot to change it on the command line...

almarklein:
Awesome, thanks @JohnLunzer !

PR #335, opened 2017-01-09
Add dialite subpackage

### LINKED ISSUE #261: A fallback GUI to communicate with the user
It would be nice to be able to communicate with the user via some sort of GUI. When Flexx is used from Python its often enough to print an error to the terminal, but in deployed apps this is not friendly and sometimes there simply is no terminal.

Possibilities:

* Something based on OpenGL. The current `flexx.lui` is a start in that direction, but should probably be removed. You'd need a lib to create the window for you, e.g. glfw, which is very small, but you'd need one for each platform that you want to support.
* Using ctypes and native calls: https://github.com/zoofIO/flexx/issues/222#issuecomment-249500840 though this would require a different implementation for Windows/OSX/Linux. Is it even possible to do on each platform?
* Using a simple web page (no JS). We're running Tornado already anyway (though what if we want to tell the user that Tornado is missing?). At least this approach should just work *anywhere* and should be relatively easy to maintain.

Native solutions:
* Windows: via ctypes using `windll.user32.MessageBoxW` but looks ugly for some reason.
* Windows: via `mshta `: `mshta "javascript:var sh=new ActiveXObject( 'WScript.Shell' ); sh.Popup( 'Message!', 10, 'Title!', 64 );close()"`
* Linux: Zenity seems pretty widely available https://en.wikipedia.org/wiki/Zenity (confirmed to even work on Raspberian)
* OSX: `osascript -e 'tell app "SystemUIServer" to display dialog "bla bla"'`

almarklein:
Closes #261

Over the holiday-ish period, I allowed myself to work on something for which there is no direct need, but which seemed like fun :)  Anyway, this provides a cross-platform, pure Python way to show simple dialogs to the users and ask questions. With only just over 500 sloc!

The most direct use-case in mind are (frozen) desktop apps that need to let the user know that a web runtime like Firefox or Chrome is needed to run the app. 

Todo:
* [x] docs
* [x] decide to put here or on separate package -> here, I want to use it
* [x] add some notes about how it works on OS X (using whatever is the frontmost app to popup a message)
* [x] maybe add retry/cancel, as downloading a resource is easy to fail e.g. if wifi is off.
* [ ] ~~would love a way to indicate progress: if a download takes a minute, the user may think something stalled.~~ I fixed this for the particular use-case in mind. Too hard to do in dialite.
* [x] proper fallback -> based on terminal, and the fallback-fallback tries to fail in a good way.
* [ ] would this work on IOS, Windows phone, Android? (probably not)

### DIFF
--- a/flexx/dialite/README.md
+++ b/flexx/dialite/README.md
@@ -1,13 +1,12 @@
 flexx.dialite subpackage
 ------------------------
 
-Dialite is a lightweight Python package for cross-platform dialogs.
-It provides a handful of functions, each one a verb, that can be
-used to inform(), warn() or fail() the user, or to confirm() something
-or ask() a yes-no question.
+Dialite is a lightweight Python package to show dialogs. It provides a
+handful of functions, each a verb, that can be used to inform(), warn()
+or fail() the user, or to confirm() something or ask() a yes-no question.
 
 Dialite is pure Python, has no dependencies, and is written to work on
-Python 3 and Python 2.x. It works on Windows (from at least XP), Linux
+Python 3 and Python 2.7. It works on Windows (from at least XP), Linux
 (most ones anyway, including Raspbian) and OS X.
 
 Dialite provides a way to communicate with the user without needing a fancy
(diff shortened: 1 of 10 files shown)


almarklein:
Maybe this should be a separate package though ...

PR #337, opened 2017-01-10
BoxPanel clarifications

almarklein:
* Add HBoxPanel and VBoxPanel to be consistent with HBox and VBox
* Clarify docs about the differences between BoxPanel and BoxLayout
* Add an example to demonstrate these differences
* Fixed that BoxPanel actuall uses `Widget.base_size` (it was hard-coded to 100).

The example app: https://dl.dropboxusercontent.com/u/1463853/differentboxes.html


almarklein:
cc @Maxyme

PR #344, opened 2017-01-27
Enable PySide2 QtWebKit runtime

jrversteegh:
The QtWebkit runtimes don't appear to work on my machine, but PySide2 should be similar to PyQt5. At least now it'll try to do something. A basis for further fixing.


almarklein:
Awesome. Thanks!

PR #345, opened 2017-01-27
Add Leaflet based MapWidget

jrversteegh:
First of all, thanks for flexx. It looks very cool (until I got Rick Roll'd by the YT widget anyway ;)
I've made a start with a map widget using leaflet. It's my first experience with flexx, so maybe it's rubbish or maybe you don't care about this kind of widget, but I thought I'd share it.
If you do care, I intend to extend it to support some more map interactions (I need to be able to at least place icons and draw lines on the map)

Hartelijke groeten, Jaap

### DIFF
--- a/flexx/util/getresource.py
+++ b/flexx/util/getresource.py
@@ -15,19 +15,22 @@ except Exception:
     warning = info = print
 
 try:
-    from urllib.request import urlopen
+    from urllib.request import urlopen, Request
 except ImportError:
     try:
-        from urllib2 import urlopen  # Legacy Python
+        from urllib2 import urlopen, Request  # Legacy Python
     except ImportError:
         raise RuntimeError('Could not import urlopen.')
 
 
 # Definition of remote resources, optionally versioned ('{}' in url becomes tag)
 phosphor_url = 'https://raw.githubusercontent.com/zoofIO/phosphor-all/{}/dist/'
+leaflet_url = 'https://unpkg.com/leaflet@{}/dist/'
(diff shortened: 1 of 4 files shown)


flexx/util/getresource.py
```diff
@@ -80,7 +83,8 @@ def _fetch_file(url):
     info('Downloading %s' % url)
     for tries in range(4):
         try:
-            return urlopen(url, timeout=5.0).read()
+            req = Request(url, headers={'User-Agent': 'flexx'})
+            return urlopen(req, timeout=5.0).read()
```
almarklein: What is this for?

almarklein:
Hoi Jaap, bedankt voor deze bijdrage :)

This is great, but I think this widget is a bit too special to be a standard widget. Also because leaflet is probably one in N ways to do maps. Therefore I'd prefer to start by making this a self-contained example, by loading the leaflet assets from CDN, similar to the jqruery and code editor examples.

Examples can still be imported by users, BTW, so it's still possible for other users to use your widget. At a later point we can then think about including it as a standard widget.

flexx/util/getresource.py
jrversteegh: For some reason https://unpkg.com/ (source of online leaflet dist) requires a user agent to be set, or it will return a 403 forbidden. Having it set, shouldn't otherwise do any harm.

jrversteegh:
OK! I'll have a look at the code-editor and jquery examples and try to move it over there.

PR #354, opened 2017-02-06
Add leaflet example

jrversteegh:
Added leaflet widget as an example.
The example 
- shows a map in a xul window
- shows how to add and remove layers
- contains checkboxes for showing the layer control and scale
- reponds to mouse clicks by adding markers to the map

I intended to make the widget reflect more of leaflet's API, but found it quite hard to come up with a generic way to do so, in particular because different applications would probably want a different level of Py <-> JS communication. I was thinking of creating a dictionary in JS with GUID->JavascriptObject entries and using the GUIDs as a reference to these objects in python, but I don't know if that's the way to go. Maybe you have some ideas on that. I am going to leave the example as it is for now.

### DIFF
--- a/flexx/ui/examples/leaflet.py
+++ b/flexx/ui/examples/leaflet.py
@@ -42,14 +42,16 @@ mimetypes.init()
 
 
 def _get_code(item):
-    """ Get a text item from _base_url """
+    """ Get a text item from _base_url 
+    """
     url = '%s/%s' % (_base_url, item)
     req = Request(url, headers={'User-Agent': 'flexx/%s' % flexx.__version__})
     return urlopen(req).read().decode()
 
 
 def _get_data(item_or_url):
-    """ Get a binary item from url or _base_url """
+    """ Get a binary item from url or _base_url 
+    """
     if '://' in item_or_url:
         url = item_or_url
     else:
@@ -59,7 +61,8 @@ def _get_data(item_or_url):
 
 
 def _embed_css_resources(css):
-    """ Replace urls in css with data urls """
+    """ Replace urls in css with data urls 
+    """
     rx = re.compile('(url\s*\(\s*(.*(\.png|\.jpg|\.svg))\s*\))')
     found = rx.findall(css)
     for match, item, ext in found:
@@ -121,7 +124,7 @@ class LeafletWidget(Widget):
         @event.prop
         def zoom(self, zoom=8):
             """ Zoom level for the map. This property is defined in Python and
-            JS because the zoomlevel can be adjust by the server as well as by
+            JS because the zoomlevel can be adjusted by the server as well as by
             the user through the map widget.
             """
             return int(zoom)
@@ -135,11 +138,14 @@ class LeafletWidget(Widget):
 
         @event.prop
         def show_layers(self, show_layers=False):
-            """ Show layers icon on the top-right of the map """
+            """ Show layers icon on the top-right of the map 
+            """
             return bool(show_layers)
 
         @event.prop
         def show_scale(self, show_scale=False):
+            """ Show scale at bottom-left of map
+            """
             return bool(show_scale)
 
     class JS:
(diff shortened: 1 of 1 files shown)


flexx/ui/examples/leaflet.py
```diff
@@ -0,0 +1,314 @@
...
+
+        def init(self):
+            with ui.BoxPanel():
+                ui.MapWidget(
+                    layers=[('http://t1.openstreetmap.org/', 'OpenStreetMap')],
+                    zoom=8,
+                    center=[52.0, 5.5]
+                )
+
+"""
```
almarklein: This example can be removed, I think its outdated.

flexx/ui/examples/leaflet.py
```diff
@@ -0,0 +1,314 @@
...
+            @event.connect('leaflet.mouse_event')
+            def handle_leaflet_mouse(self, *events):
+                ev = events[-1]
+                latlng = tuple(ev['latlng'])
+                if ev['event'] == 'click':
+                    m = L.marker(ev['latlng'])
+                    m.bindTooltip('%f, %f' % (latlng[0], latlng[1]))
+                    m.addTo(self.leaflet.map)
+
+    app.launch(MapWidget, 'xul')
```
almarklein: Just a heads up that I am refactoring the webruntime module, and this should soon be "app" instead (or "firefox-app" if you want to use Xul per see).

almarklein:
This is great. I think we should merge this as soon as that one comment is covered.

In the mean time, in light of maybe a `flexx.ui.MapWidget` how do you think that leaflet compares to other frameworks such as those listed here: http://wiki.openstreetmap.org/wiki/Frameworks?

flexx/ui/examples/leaflet.py
jrversteegh: I noticed. There are a few more examples that use "xul", so I guess that will have to be fixed anyway after you merge.

jrversteegh:
Depends a bit on the goal imho. It seems current development leans towards rendering vector maps with WebGL, which makes sense, but I don't have the impression it's really there yet.
For 'simple' tile maps, the only reasonable options appear to be openlayers, leaflet and Google maps. See e.g. http://wiki.openstreetmap.org/wiki/Deploying_your_own_Slippy_Map
I haven't really looked into google maps, but between openlayers and leaflet I found in my little research:
 - openlayers is GIS oriented with an extensive API, fairly steep learning curve, and large code base (~1M)
 - leaflet is web oriented, focused on simplicity, extensible with plugins, easier to get started with and smaller (~150K)
As a general advice I found: if you want to build an extensive mapping application, take the openlayers pill, if you just want to fling together some mapping stuff, leaflet is your best friend. Opinions differ: http://gis.stackexchange.com/questions/33918/should-i-use-openlayers-or-leaflet

flexx/ui/examples/leaflet.py
jrversteegh: removed.

jrversteegh:
For a vector based approach, we might have a look at https://github.com/mapbox/mapbox-gl-js

flexx/ui/examples/leaflet.py
almarklein: I'm changing those in the same PR, but with a bit of luck I'll remember to rebase after this is merged :)  and if not, 'xul'  will still work, but show a warning.

almarklein:
> It seems current development leans towards rendering vector maps with WebGL, which makes sense, but I don't have the impression it's really there yet.

I can imagine that geometry shaders that come with WebGL2 will make this more feasible.

almarklein:
Thanks for the awesome contribution!

jrversteegh:
You´re very welcome. I hope to able to contribute some more in the future. If you have a specific task that you would like me to take on. Let me know.

jrversteegh:
> I can imagine that geometry shaders that come with WebGL2 will make this more feasible.

Ah. You probably know more about the details of the actual rendering than I do. I don't know if technical feasibility is still an issue. I meant "ready" in the sense of e.g. availability and adoption of a standard for http://wiki.openstreetmap.org/wiki/Vector_tiles and stuff like that.

PR #365, opened 2017-03-09
Cookies!

almarklein:
This PR adds functionality to the `Session` class to set and get (encrypted) cookies. Also adds an example to demonstrate the use of cookies.

Note that cookies don't really work for desktop apps, and when launching an app with `app.launch()`, the cookies only become available when the websocket is set, which is *after* the main model initializes. For served web apps (which I expect to be the only real use-case for this feature) this is not a problem.


almarklein:
cc @Maxyme

Issue #364, opened 2017-02-24
Adopt a Flux-like approach to let information flow through an application


almarklein:
This is a proposal for some pretty profound changes to how building apps with Flexx works. Though Flexx is still marked alpha, there are people using Flexx already, and they'd need to make some serious changes if we go this route. We thus have to think carefully whether the proposed changes are worth it.

## Background

We've found Flexx to be working pretty well, even for larger applications, though the complexity of an app increases too, and we've found that newcomers to our app have trouble understanding the data-flows etc. I feel that #359 is a symptom of this, and that there is a deeper cause.

Flexx uses events. A lot. Everything is tied together using events. This is the case for both user  input (mouse and key events) and property changes. The system works very nice (if I may say so myself), but a problem is that information flows all over the place, which makes it difficult to reason about the state of the application.

## Flux and friends

At Facebook they had similar problems with their classic MCV approach; it scales badly because its hard to predict the state of an application based on stuff that happens in the app. Their core insight was that if information flows in one direction, an app becomes much more predictable. Here is a [video](https://youtu.be/nYkdrAPrdcw) that explains it very clearly (thanks @korijn) and a [comic](https://code-cartoons.com/a-cartoon-guide-to-flux-6157355ab207).

At Facebook they came up with a pattern (and framework)  that they call Flux. The same pattern has been used in different places, in different variants, e.g. Redux and Veux. The core principal is that information flows like this:
```
Actions --> Dispatcher  --> One ore more stores --> Views
                /\                                    |
                 - - - - - - - - - -  Actions - - - - -
```

The store(s) represent the state of the application. The views listen to changes in the state and update the UI accordingly. Views (and other parts of an app) can create actions (kind of events) that are handled by the store and eventually give rise to changes in the state.

An important concept is that any actions that are dispatched (e.g. from views) are only handled *after* the view is done processing stuff. That's why we can legitimate saying that data flows in one direction, even though we see an arrow going to the left. Also important is that updates to the store are atomic, "commiting mutations" in Veux-terminology. The store handles actions, and then either applies mutations, or it may e.g. call out to an API, schedule some work, and eventually (asynchronously) apply mutations. Or both. This [page](https://vuex.vuejs.org/en/mutations.html) explains it in more detail.

The state in the store is a higher-level representation of the application state; instead of having a label for which its text property represents the username, the app would have a "username" entry in the store, and a view would be responsible for showing it. 

## Adopting this pattern in Flexx

[...]

PR #367, opened 2017-03-16
Proposal of changes to make Flexx scale better

almarklein:
Here is a preliminary proposal of changes to make writing (complex) apps easier and make Flexx scale better. Everything is up for discussion, so don't hesitate to challenge it.

For context and discussion see #364. I am using a PR for this proposal, so that others can easily react in-line with the text. See the "files changed" section for what I think the docs for the new Model class could look like. Below I give a summary with argumentation.

----

## Introduction

The problems that I'd like to address with this proposal:

* A) Hard to predict flow of information in general.
* B) Need rather verbose code for simpler handlers and properties.
* C) Confusion of user on how the current `Model` works. Py and JS is too "entangled".
* D) Flickering effect of properties (in Python) due to eventual synchronicity.
* E) Complex memory management (it was quite leaky until #380, also see #381).
* F) Its not trivial to write apps with centralized storage.
* G) No story for updating array data effectively (e.g. scientific purposes).
* H) 

First off, in my view, the Elm architecture (Flux, Veux et al.) provides two main advantages/insights: having global state that can be observed/used in different parts of the application (helping with F); having a predictable flow of information, where certain user-interactions in the view can cause an "action" that starts a new flow of information (helping with A and D).

It took me a while to come up with an approach to include these ideas in Flexx, while also keeping things simple. I've considered everything from tiny changes to a complete overhaul of the event system. The latter is not what I propose, though some changes are pretty profound and not backward compatible. 

## Summary + argumentation

action state reaction

### Components
* Rename Model to Component, because it takes the role of both the M and V in MVC. The new component comes in two flavors: 
* The `PyComponent` has all its logic in Python. It is instantiated in Python and can even be associated with multiple clients (e..g. for shared data in a chat app).
* The `JSComponent` has all its logic in JavaScript. It can be instantiated from Python though, in which case its state can be observed and actions can be invoked.
* No more need for the nested classes.
* The flow of information starts at actions, which mutate state, upon which there are reactions. Actions and reactions are really just methods on a Component class.
* State (properties) are readonly (makes code predictable) and can only be mutated from actions.
* Widgets should be more optional; it should be easy (and not feel like a hack) to use other means to realize the view of an app, e.g. using bootstrap.
* All components have a `root` attribute that refers to the main application component. This can serve as a place to keep global state (a "store" in Veux/Flux terminology).

### Properties (state)
* Properties can be defined on a single line (saves lines of code).
* A standard (typed) `set_foo()` action can be created automatically, because its so common by `foo = prop(settable=int)`.
* Introduce a "computed property", similar to automagic reactions mentioned below.
* Introduce an `array_prop` to manage state in an array. Elements can be heterogeneous (list) or typed arrays (ctypes arrays in Python). Mutations can append/insert/remove items, making syncing/tracking such state very efficient.
* Perhaps introduce a `dict_prop` which can similarly be mutated partially.

### Actions
* Calling an action runs the action async (i.e. later), unless its called from another action. This makes it that the flow of information "starts" at the action.
* Actions can be called (i.e. invoked) from both Python and JS, as long as the args can be serialized. Think of `button.set_text()`.
* Flexx handles one action at a time, and all reactions caused by the mutations done by the action are handled *before* moving on to the next action. 

### Reactions
* Reactions are more or less what we call handlers now, using `@connect(...)`.
* Reactions react to mutations to make the app reflect the new state.
* Reactions react to events, in most cases by invoking new actions.
* This still works: `@reaction(... connection strings ...)`
* We also allow just `@reaction`, and let Flexx figure out the dependencies automagically.
* Inline reactions: `Label(text=lambda:self.username)` and `Button(on_mouse_click=lambda ev: self.submit())`. These will greatly help reduce code.

### Renaming / restructuring

* `Model` ->`PyComponent` and `JSComponent`.
* The `events` subpackage made sense since Flexx is now very event-centric. The proposed changes make it more state-centric, and the name bugs me. Perhaps I'll just move it all to `flexx.app`.
* Maybe spin out `pyscript` and `webruntime`, so that we can just do `flexx.PyComponent`?

### Similarity and differences w.r.t. Elm/Flux/Veux/Redux

Redux adopts a very "functional" approach to dealing with state, where reducers (i.e. actions) receive the state as an input argument, and return a new state. I don't doubt that this helps predictability, but it means having to make a copy of the whole state at each transition. This breaks down if we want to do something with data. It also does not work well with the OO style that Flexx uses.

A difference with Flux/Veux is that what they call "view", we call "reactions". One reason is that our reactions will often involve invoking actions on lower level components (e.g. widgets) rather than do DOM updates directly. So the scope of our reactions is somewhat broader, but the idea is very similar.

In most of the aforementioned frameworks there is a concept of a global/centralized state object, though each framework has another means to have some kind of modularity. In Flexx each component has its own state, and its up to the user to design the relationships in a sensible way, e.g. defining a widget's state as "local state" and using a common component, or the root component as "application state".

### Debugging / testing

The clear flow of information also makes is possible/easier to create fancy debugging tools. E.g. see the list of pending actions, clicking a button to process the first one, see what mutations it causes, and which reactions follow from it. Similarly, it might help with (unit) testing.

### Separation of Python and JS

It's become clear to me that Flexx' power to bring Python (server side) code and JavaScript (client side) code close together is also its weakness; it is confusing for users.

My initial take (written down originally in this PR) was to allow instantiating components in JS, in which case they exist/operate only in JS. That would make it easier to write apps that clearly separate JS code from Py code (i.e. views from application logic). I now think that this is not sufficient.

By making a component either operate in JS or Python, the concerns of such a component are much more clear. The state that can still be observed from the other end, and the actions that can be invoked, provide an easy API to handle the Python-JS interactions.

A point of discussion was whether js-components could be instantiated from Python. If not, we'd need a way to somehow bind a Python and JS entry point together, which would greatly reduce the ease and number of places where JS and Python can interact. IMO the possibility to instantiate and *use* a js-component from Python does not break the JS/Py separation too much, since its still fully *implemented* in Python.

We'd need PyScript to support context managers and keyword args to allow instantiating components in JS in the same way as we do now.

### array_prop

I'd like a better story to manage arrays as state, without having to resend the whole array on each change, and making it easy for a view to update its local representation without a hard reset. Thinking of WebGL in particular, but also e.g. a list view with a lot of items. On top of this, one could implement "data sources" e.g. to manage image data.

To send mutations with data in the form of typed arrays we need a way to send binary messages. To keep things in order, this must happen over the same single websocket. So we should probably use a binary protocol instead.

### DIFF
--- a/new/colab_painting.py
+++ b/new/colab_painting.py
@@ -2,10 +2,10 @@
 New version of the colaborative paiting example. The PyModel makes this so simple!
 """
 
-from flexx import app, ui
+import flexx.ui
 
 
-class SharedState(app.PyModel):
+class SharedState(flexx.PyComponent):
     
     points = app.array_prop()
     
@@ -19,27 +19,29 @@ class SharedState(app.PyModel):
 shared = SharedState()
 
 
-class ColabPainting(app.Widget):
+class ColabPainting(flexx.PyComponent):
     
     def init(self):
+        self.view = ColabWidget(shared)
+
+
+class ColabWidget(flexx.JSComponent):  # ----- This runs in JS
+
+    def init(self, shared):
         self.shared = shared
+        self.canvas = flexx.ui.CanvasWidget()
     
-    class JS:
-        
-        def init(self):
-            self.canvas = ui.CanvasWidget()
-        
-        @app.reaction('canvas.mouse_down')
-        def _mouse_down(self, *events):
-            for ev in events:
-                self.shared.add_point(ev.x, ev.y, 'red')  # invoke action
-    
-        @app.reaction
-        def draw(self):
-            ctx = self.canvas.node.getcontext('2d')
-            for x, y, color in self.shared.points:
-                ctx.fillStyle = color
-                ctx.fillRect(x, y, 5, 5)
+    @app.reaction('canvas.mouse_down')
+    def _mouse_down(self, *events):
+        for ev in events:
+            self.shared.add_point(ev.x, ev.y, 'red')  # invoke action
+
+    @app.reaction
+    def draw(self):
+        ctx = self.canvas.node.getcontext('2d')
+        for x, y, color in self.shared.points:
+            ctx.fillStyle = color
+            ctx.fillRect(x, y, 5, 5)
         
         # todo: when there has been no resizing of the canvas, you could even draw just the newly added points
 
--- a/new/colab_painting.py
+++ b/new/colab_painting.py
@@ -1,0 +1,49 @@
+"""
+New version of the colaborative paiting example. The PyModel makes this so simple!
+"""
+
+from flexx import app, ui
+
+
+class SharedState(app.PyModel):
+    
+    points = app.array_prop()
+    
+    @app.action
+    def add_point(self, x, y, color):
+        self._push_prop('points', (x, y, color))
+        too_much = len(self.points) - 10000
+        if too_much > 0:
+            self._remove_prop('points', 0, too_much)
+
+shared = SharedState()
+
+
+class ColabPainting(app.Widget):
+    
+    def init(self):
+        self.shared = shared
+    
+    class JS:
+        
+        def init(self):
+            self.canvas = ui.CanvasWidget()
+        
+        @app.reaction('canvas.mouse_down')
+        def _mouse_down(self, *events):
+            for ev in events:
+                self.shared.add_point(ev.x, ev.y, 'red')  # invoke action
+    
+        @app.reaction
+        def draw(self):
+            ctx = self.canvas.node.getcontext('2d')
+            for x, y, color in self.shared.points:
+                ctx.fillStyle = color
+                ctx.fillRect(x, y, 5, 5)
+        
+        # todo: when there has been no resizing of the canvas, you could even draw just the newly added points
+
+
+if __name__ == '__main__':
+    app.serve(ColabPainting)
+    app.run()
--- a/new/shopping_cart.py
+++ b/new/shopping_cart.py
@@ -11,7 +11,7 @@ class that handles the logic and also does the rendering.
 from flexx import app, ui
 
 
-class Store(app.PyModel):
+class Store(app.PyComponent):
     """ A server-side model that tracks the inventory at the server. Shared by all
     clients.
     """
@@ -45,7 +45,7 @@ class Store(app.PyModel):
 store = Store()
 
 
-class Cart(app.Model):
+class Cart(app.JSComponent):
     """ Representation of (the bisiness logic of) a shopping cart.
     """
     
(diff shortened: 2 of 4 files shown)


almarklein:
flexx info flow

Here's a more detailed diagram of the flow of information in Flexx. Actions are queued (similar to events) and handled one action at a time. After processing an action, the state might have been changed, which has
caused the queue of reactions to fill up. These are handled next until there are no more reactions. Only then is the next action processed. When processing reactions, the state is guaranteed to stay the same.

Reactions to events (e.g. user interactions) are queued in a separate queue and processed separately. In most cases such reactions will simply invoke a new action. Though it can also cause low-level changes in the view (not related to state, like animations).

If we put Python in the mix, we get a slightly bigger diagram. The bottom half is nearly a mirror image, except that it cannot mutate state (in needs to invoke actions to affect the state).

flexx info flow +

almarklein:
Added more examples (as files in the PR). Let me know if you'd like to see a certain use-case.

almarklein:
After discussions with @korijn and several Flexx "newbies", some insights/changes w.r.t. this proposal:
* The term "model" is confusing because a `Model` is used to represent both models and views. 
* Users find it hard to distinguish between Python land and JS land. This is partly due to wrong expectations, but it does not help that you cannot use numpy in a module that says `import numpy` at the top (as one user put it).
* My current stance is to implement subclasses of `JSPart` and `PyPart`, which are either fully implemented in Python or fully implemented in JS. This e.g. makes it easy for a large app to put all `PyPart` classes in one module/subpackage, and all `PyJS` classes in another. `JSPart` classes can be instantiated in Python, so that its state can be observed, and its actions invoked. Yet all mutations/logic occurs in JS.

almarklein:
Updated the description and new Model/Component docs to reflect the idea of having dedicated JSComponent and PyComponent classes.

Some thought about the mutator functions, which are probably hard to remember in the current proposal. Maybe make it more Pythonic:

```py
@action
def update_foo():
    
    # A: Current proposal
    self._set_prop('foo', [1, 2, 3])
    self._push_prop('foo', index, ['new', 'elements'])
    self._update_prop('foo', index, ['updated', 'elements'])
    self._remove_prop('foo', index, n)
    
    # B: Maybe this?
    self._mutate_foo('set', [1, 2, 3])
    self._mutate_foo('push', ['new', 'elements'])
    self._mutate_foo('insert', index, ['new', 'elements'])
    self._mutate_foo('update', index, ['updated', 'elements'])
    self._mutate_foo('remove', index, n)
    
    # C: More Pythonic, but maybe too much?
    self.foo = 1, 2, 3
    self.foo.appendn('new')
    self.foo.extend(['new', 'elements'])
    self.foo.insert(index, 'new')  # but how to insert multiple?
    self.foo[index] = 'updated'
    self.foo[index:index+n] = 'updated', 'elements'
    self.foo.pop(index, n)
```

In C, `foo` is set as if it were a property. Note that this would only be allowed inside an action. Reactions need to use `set_foo()` or whatever action is available to modify `foo`. Maybe this makes it too easy? The idea was that the actions are responsible for validating the input. If its so easy to set a property, its tempting to set the prop of another component, and then the responsibility of validation is spread all over the place. Also, what about read-only properties? 

With private mutator functions each components simply provides the actions it wants to let other components influence its state. This is why I now lean towards option B.

niki-sp:
On 30.03.2017 14:30, Almar Klein wrote:
> Updated the description and new Model/Component docs to reflect the idea
> of having dedicated JSComponent and PyComponent classes.
>
> Some thought about the mutator functions, which are probably hard to
> remember in the current proposal. Maybe make it more Pythonic:
>
> @action
> def update_foo():
>
>     # A: Current proposal
>     self._set_prop('foo', [1, 2, 3])
>     self._push_prop('foo', index, ['new', 'elements'])
>     self._update_prop('foo', index, ['updated', 'elements'])
>     self._remove_prop('foo', index, n)

This variant is easy to google - _set_prop() is found in many projects 
and easy to document.
>
>     # B: Maybe this?
>     self._mutate_foo('set', [1, 2, 3])
>     self._mutate_foo('push', ['new', 'elements'])
>     self._mutate_foo('insert', index, ['new', 'elements'])
>     self._mutate_foo('update', index, ['updated', 'elements'])
>     self._mutate_foo('remove', index, n)

_muttate_foo() is project specific ans hard to google. Also function as 
first argument it error prone and IMHO unpythonic.
>
>     # C: More Pythonic, but maybe too much?
>     self.foo = 1, 2, 3
>     self.foo.appendn('new')
>     self.foo.extend(['new', 'elements'])
>     self.foo.insert(index, 'new')  # but how to insert multiple?
>     self.foo[index] = 'updated'
>     self.foo[index:index+n] = 'updated', 'elements'
>     self.foo.pop(index, n)

This looks somewhat similar to PEAK toolkit.

-1 for variant B.

Niki

almarklein:
```
# D: easy for basic setters, same function for "advanced" mutations
self._set_foo([1, 2, 3])
self._set_foo('push', ['new', 'elements'])
self._set_foo('insert', index, ['new', 'elements'])
self._set_foo('update', index, ['updated', 'elements'])
self._set_foo('remove', index, n)
```

niki-sp:
On  6.04.2017 13:51, Almar Klein wrote:
> |# D: easy for basic setters, same function for "advanced" mutations
> self._set_foo([1, 2, 3])
 > self._set_foo('push', ['new', 'elements'])
> self._set_foo('insert', index, ['new', 'elements'])
> self._set_foo('update', index, ['updated', 'elements'])
> self._set_foo('remove', index, n) |

variant:

self._set_foo(['new', 'elements'], 'push') # use insert=sys.maxint?
self._set_foo(['new', 'elements'], insert=index)
self._set_foo(['updated', 'elements'], update=index)
self._set_foo([], update=(index, n))
self._set_foo(remove=(index, n))

insert=index === update=(index, 0)

Niki

frmdstryr:
These don't actually work yet, just a proposal correct?

> Traceback (most recent call last):
>   File "flexx/new/shopping_cart.py", line 14, in <module>
>     class Store(app.PyComponent):
> AttributeError: 'module' object has no attribute 'PyComponent'

almarklein:
> These don't actually work yet, just a proposal correct?

Yes, this is still only a proposal.

almarklein:
```py
    # E
    self._set_foo([1, 2, 3])
    self._push_foo(['new', 'elements'], index)  # omitting index adds to end
    self._pop_foo(n, index)  # omitting index pops from start
    self._update_foo(['updated', 'elements'], index)
```

Korijn:
Here is my take on this design pattern: https://github.com/Korijn/plux

almarklein:
Some comments after discussions and subsequent reflections.

This proposal addresses multiple problems. I listed these in the initial post of this issue. Being able to use a central store when developing apps is one of these problems. But the Elm-approach does more.

I argue that the Elm-approach helps with other issues too. In most MVC apps, components observe each-other and react directly. This can quickly result in complex dependency diagrams and hard to predict behavior. In the presence of loops, the app can break. When writing the current approach of Flexx, I foresaw this and and made Flexx components process the reactions in the next event loop iteration. In other words, *indirectly*. However, this makes it harder to track the cause of a certain reaction, and the order in which reactions are processed is not well defined, because the reaction is queued on top of the event queue, and wont be processed until several other events are processed.

The current proposal solves this by queuing actions (in the order that they were invoked) and by processing all reactions to state mutations before moving on to the next action. 

----

A shortcoming of the current proposal, compared to Elm et al. is that by using predefined mutations, we do not have semantic meaning for the mutations. In fact, what are called "mutator functions" methods here is just API to modify the state. So I guess its fair to say that the current proposal does not use mutations.

Mutations in Flux et al. are, as I know understand it, intended to group multiple changes in state into one atomic unit that can be considered one "item" in the state history, and to associate semantic meaning to that state change,

I'd love to put that grouping and semantic in there somehow. I imagine it would make debugging a lot nicer. Perhaps via something like:
```py
@action
def reset_items(self):
    with self.mutate('reset items'):
        self._set_items([])
        self._set_count(0)
```

JohnLunzer:
I am curious about the status of Flexx development. I know such a big refactoring would have to be done "behind the scenes", but from GitHub it appears that Flexx development has stalled. There are probably a lot of us that would be happy to hear how things are going.

Korijn:
Almar has been on a three week holiday and is also attending a python conference, so he's been absent for a little over a month.

almarklein:
I'm back to work, but a refactoring like this needs a good "time slot" to work on it; its too complex to do "on the side".

almarklein:
In a recent conversation @ivoflipse wondered about performance when a certain event occurs in fast succession over a period of time. In the current implementation event handlers get a list of events and can chose to only react to the last one. In the proposal, this still applies to reactions (e.g. mouse_move handler), but not to actions.  See https://github.com/zoofIO/flexx/pull/367#issuecomment-287050936

In the post where I posted that diagram I mentioned that actions are handled one at a time, and all reactions are processed before the next action is handled. I now think that this is not necessary. It *is* important to handle all reactions before processing any action again, but I think that we can actually flush the whole queue of actions (or part of it, really) before handling any reactions, which allows for more  "event skipping".

Another issue that came up in the same discussion was about propagation of state: if one e.g. has a property that represents a list of items, and this list is visualized using a TreeWidget (by reaction to the former property), then there are moments when the reference property is updated, but the state of the tree not yet. This, is however, a somewhat intrinsic feature of event based systems. The state of the tree widget should be considered derived state and should probably not be reacted to if one is really interested in the original property. That said, there might be cases where derived state *is* important, and in this proposal, one can make the action that sets the original property also set the derived state: reactions are then guaranteed to always work with a consistent state. (In the current implementation this can be achieved by setting the derived property in the property setter of the reference property.)

PR #383, opened 2017-06-14
add Session.call_after_roundtrip

almarklein:
Adds a convenience method on `Session` to overcome jitter effects. We keep patching things up, but really, this idea of objects that exist in both Python and JS is flawed.


Korijn:
This is purely a workaround, no?

almarklein:
You could consider this a feature to deal with (i.e. work around) some of the limitations of the current `Model` mechanics (in particular the way how properties are synced).

PR #389, opened 2017-07-21
TreeWidget multiselect with Ctrl and Shift

almarklein:
* `TreeItem.mouse_click` and `TreeItem.mouse_double_click` events now carry more information about buttons and modifier keys (needed to implement the below).
* When `max_selected` is `<1`, one can use `Ctrl` and `Shift` to select multiple items.
* To get the old behavior, use `max_selected=1e9`.
* Doubleclicking an item selects it (previously it would just toggle on and off (or off and on).

Also fixed issues with Travis CI / Pytest / Pypy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,10 +20,10 @@ matrix:
         #
         - python: "2.7"
           env: TEST_UNIT=1 TEST_INSTALL=1
-        - python: "pypy-5.8"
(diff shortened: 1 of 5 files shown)


almarklein:
Seems that Travis is updating from Ubuntu Precise to Trusty, and we ran into https://github.com/travis-ci/travis-ci/issues/6865

PR #392, opened 2017-08-18
Added missing return statement

ivoflipse:
This fixes a problem where double clicking events weren't emitted


almarklein:
Eek, ugly how that slipped through! Thanks!

Issue #393, opened 2017-08-19
Tornado debug mode


gregfreeman:
I'd like to use the tornado debug mode for development.  

This gives the ability to autoreload the flexx app when developing. 

I'll send a PR

What do you think?

almarklein:
Closed by #394

PR #394, opened 2017-08-19
add config tornado_debug to allow running tornado in debug mode

gregfreeman:
Here's a pull request for  issue #393 that I just created.   

I added a config attribute to set tornado debug mode.   

I tried some test scripts for this, but it was a little tricky reloading the app without reloading the test.

### DIFF
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -25,4 +25,7 @@ config = Config('flexx', '~appdata/.flexx.cfg',
                     'Auto-detect by default.'),
         nw_exe=('', str, 'The location of the NW.js executable. '
                 'Auto-install by default.'),
+        # tornado
+        tornado_debug=('false', bool, 'Setting the tornado application debug flag '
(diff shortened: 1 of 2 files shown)


almarklein:
This sounds like it can be useful. Since all this does is pass an argument to Tornado, and is a feature only used by specific users (devs), let's assume that Tornado tested this feature and "manually test" whether it works with Flexx.

almarklein:
Travis seems to be a bit glitchy lately, but it turned green after rerunning.

almarklein:
Is this ready from your end?

gregfreeman:
Yes

On Aug 23, 2017 05:33, "Almar Klein" <notifications@github.com> wrote:

Is this ready from your end?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/zoofIO/flexx/pull/394#issuecomment-324289675>, or mute
the thread

almarklein:
Thanks!

Issue #189, opened 2016-08-22
Routing and linking subsystem


Korijn:
Something that Flexx currently lacks in comparison to other UI toolkits is linking and routing.

An app with pages, links and URLs is typically implemented with a Router for single page applications. This usually also syncs with browser history to ensure the Back button functions, and so that users can share links. An example of this is react-router:(https://github.com/reactjs/react-router/blob/master/README.md).

What work do you think would be required for such a subsystem?

almarklein:
I can imagine an app where you host multiple apps and have links between them. Would that give you what you mean? Maybe Flexx needs a way to write something like that (or the links) down more effectively?

Korijn:
Well, as mentioned earlier, I think this is closely related to #182. Switching out whole hierarchies of subwidgets would be the main purpose of a `Router` widget. Triggering this can be done programmatically, and a `Link` widget would be designed ready-to-go.

Side-effects of the `Router` would include URL and history tracking.

IMHO, linking between multiple apps seems unrelated and more like something you can accomplish by implementing e.g. #178.

almarklein:
Summarising in my own words: this feature would be a per-app ability to map the remainder of the url (after the app name) to the "configuration" of the app itself. E.g. `http://foo.com/users` would show a list of users, and `http://foo.com/users/23` would show user 23.

Korijn:
A nice example can be found here: https://github.com/reactjs/react-router/blob/master/examples/active-links/app.js

PR #398, opened 2017-09-07
Insert the path into the session when creating it

Konubinix:
It follows the same logic as setting the cookies: it is added to
the object right after it has been constructed and makes sense
only in the scope of web applications.

### DIFF
--- a/flexx/app/_session.py
+++ b/flexx/app/_session.py
@@ -62,10 +62,10 @@ class Session:
 
     STATUS = new_type('Enum', (), {'PENDING': 1, 'CONNECTED': 2, 'CLOSED': 0})
 
-    def __init__(self, app_name, store=None):  # Allow custom store for testing
+    def __init__(self, app_name, store=None,
+                 request=None):  # Allow custom store for testing
         self._store = store if (store is not None) else assetstore
         assert isinstance(self._store, AssetStore)
-        self._cookies = {}
 
         self._creation_time = time.time()  # used by app manager
 
@@ -100,8 +100,13 @@ class Session:
         # commands, which are send to the client as soon as it connects
         self._pending_commands = []
 
-        # record the request that originated the session
-        self._request = None
+        # request related information
+        self._request = request
+        if request and request.cookies:
+            cookies = request.cookies
+        else:
+            cookies = {}
+        self._set_cookies(cookies)
 
     def __repr__(self):
         t = '<%s for %r (%i) at 0x%x>'
--- a/flexx/app/_session.py
+++ b/flexx/app/_session.py
@@ -25,11 +25,11 @@ reprs = json.dumps
 
 # Use the system PRNG for session id generation (if possible)
 # NOTE: secure random string generation implementation is adapted
-#       from the Django project. 
+#       from the Django project.
 
 def get_random_string(length=24, allowed_chars=None):
     """ Produce a securely generated random string.
-    
+
     With a length of 12 with the a-z, A-Z, 0-9 character set returns
     a 71-bit value. log_2((26+26+10)^12) =~ 71 bits
     """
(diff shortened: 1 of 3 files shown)


Konubinix:
Please tell me if you prefer my to submit a patch that preserve trailing whitespaces. I configured my text editor to automatically remove trailing whitespaces and though it would not hurt.

almarklein:
Can you provide a use-case / small code sample how you'd use this feature?

> Please tell me if you prefer my to submit a patch that preserve trailing whitespaces

That's ok, thanks to [?w=1](https://github.com/zoofIO/flexx/pull/398/files?w=1)

Konubinix:
I created a file explorer with flexx and I wanted to encode into the url
the path to open.

http://...../FileExplorer/some/path/to/my/files

Till now, I just asked the javascript part to provide the python part with
the path.

```python
#!/usr/bin/env python3
# -*- coding:utf-8 -*-

from flexx import app, ui, event


class FileExplorer(ui.Widget):
    def init(self):
        self.path_widget = ui.Label()
        self.emit("init_js", {})

    @event.connect("thepath")
    def on_thepath(self, *evs):
        self.path_widget.text = "The path is {}".format(evs[-1]["value"])

    class JS:
        @event.connect("init_js")
        def on_init_js(self, v):
            self.emit(
                "thepath", {
                    "value": window.location.pathname[len(flexx.app_name) + 2:]
                }
            )


if __name__ == "__main__":
    app.serve(FileExplorer)
    app.start()
```

I wish to reduce this to simply
```python
#!/usr/bin/env python3
# -*- coding:utf-8 -*-

from flexx import app, ui, event


class FileExplorer(ui.Widget):
    def init(self):
        self.path_widget = ui.Label(text="The path is {}".format(self.session.path))
        self.emit("init_js", {})


if __name__ == "__main__":
    app.serve(FileExplorer)
    app.start()
```

Konubinix:
By the way. I am not sure how to do the first example using event.prop instead of sending events...

almarklein:
Well, only the browser knows that path, so if you want Python to know about, it you'll have to communicate it :)

Your JS init code can be in the function called `init()` and be called automatically at startup. You can make a synced property to easily communicate the value:
````py
...
class FileExplorer(ui.Widget):
    ...
    class Both:
        @event.prop
        def the_url(self, v):
            return str(v)

    class JS:
        ...
```

Konubinix:
This is not true. But maybe I was not clear.

When the browser performs the HTTP GET request (/FileExplorer/some/path), the
web server (tornado) analyses this path to decide to which handler it should
provide the request. In our case, the request is handled by AppHandler.

AppHandler, in get and then in _get_app, performs some analyses and
transformation of the full_path. One of those is to extract the so called path
(some/path), by removing the application name (FileExplore) from the beginning
of the full_path.

This is were the path is lost: A session is created, the page is returned to
the browser and the rest of the communication takes place in the websocket.

Therefore, it is not true that only the browser knows the path. Tornado and the
AppHandler know it at some point. They just don't communicate it to the python
Model. I though that the session was the good place to give such information.

Thank you for the information about the init method in the JS class and for the
suggestion to use a synced property in Both. For some reason, I thought I could
do that by writing a property in the python part and set it in the JS part.

Korijn:
> For some reason, I thought I could
do that by writing a property in the python part and set it in the JS part.

That could be a very interesting way to make that explicit though. @almarklein `self.py.url = window.location.pathname[len(flexx.app_name) + 2:]`

almarklein:
Ah, I see what you mean now. I guess this is more or less what is proposed in #189.

Konubinix:
@almarklein I think that #189 is way more ambitious than my proposal.

While I suggest to provide the path to the application and let it do whatever it
wants with it by parsing it, #189 proposes to already pre process the path in a
routing mechanism so that the application can get arguments ready to be
consumed.

I think that if #189 was available, I would not need to get the path, and that
would be awesome. Nonetheless, it is very likely that such a feature will be
part of flexx before long. Therefore, I consider providing the path as-is to the
application a small win now waiting for a big win in the future :-).

almarklein:
That's fair enough, and I imagine that it won't get in the way of #189, since it'll be ok to have a path as well as parsed options. So what's still needed to make this work?

Konubinix:
AFAICU, this is sufficient to fit my needs. But, now that I am thinking and comparing what other framework do, I think that it might be better to store the whole request in the session instead of the path.

Indeed, IIRC, in all the web frameworks I tried so far, the handler had access to the pre computer path like suggested in #189 as well as the bare request.

Also, I'd like to get access to the path today, but I may want to get the arguments in the future, or the sent cookies/headers. Like in the case of letting the path through, I think that it does not hurt to provide the whole request and it will most likely suit a handful of needs. Actually, when I tried to find out how to access the path, one of my first attempts was to check whether I could do self.session.request.path :-).

If you don't mind, I will change the pull request to provide the request instead of the path into the session.

Konubinix:
Also, since the request contains the cookies, I can make create_session take request as parameter instead of cookies.

Konubinix:
I changed the commit to provide the request instead of the path. Was do you think about this?

Konubinix:
With the request, you can also call the application with 

```
/FileExplorer/?path=/some/path
```
And get the path in the model with self.session.request.arguments["path"]

almarklein:
This looks nice; sending along the request seems like it might be useful indeed.

> `session.request = request`

Could you add a `@property `for `request` so its documented? Maybe do `session = Session(name, request=request)` so that the session can set the request (and cookies) itself.

Konubinix:
Done :-)

Konubinix:
I fixed the style error in session.

I am not sure the Sphinx error is related to my changes. Can you tell me what is going on with Sphinx ?

Konubinix:
Nevermind. No sphinx error today :-)

flexx/app/_session.py
```diff
@@ -50,89 +50,102 @@ class Session:
...
+
     def __repr__(self):
         t = '<%s for %r (%i) at 0x%x>'
         return t % (self.__class__.__name__, self.app_name, self.status, id(self))
-    
+
+    @property
+    def request(self):
+        """The tornado request that was at the origin of this session
+        """
```
almarklein: `return self._request`

Konubinix:
Oups.... It should be better that way :-)

almarklein:
Looking good. That one CI runner is a glitch. Trying to do something about it in another PR. Thanks!

PR #400, opened 2017-09-12
Added config option to disable stack trace in browser window

jrversteegh:
On a production server it's not desirable to show users a stack trace in case of an error. Depending on the audience this might be for user friendliness or security reasons. I feel the option should default to False, but set to True for now to maintain current behaviour.

### DIFF
--- a/flexx/_config.py
+++ b/flexx/_config.py
@@ -4,6 +4,7 @@ config = Config('flexx', '~appdata/.flexx.cfg',
 
         # General
         log_level=('info', str, 'The log level to use (DEBUG, INFO, WARNING, ERROR)'),
+        browser_stacktrace=(True, bool, 'Show stack traces in browser window'),
         
         # flexx.app
         hostname=('localhost', str, 'The default hostname to serve apps.'),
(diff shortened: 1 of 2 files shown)


almarklein:
Can you expand on why this is useful?

This applies to the reporting of server errors when loading a page. IMHO I don't see how hiding these would help. The end-user then knows that something is wrong instead of getting only a status code, and can even copy-paste the error message when reporting the problem, which should help the dev team to fix the problem.

jrversteegh:
In my experience even fairly technical users like engineers don't appreciate a stack trace. If there is an error, they'll just report this fact. It even takes considerable effort to get them to post the actual error message in an issue report. The stack trace itself can be found in the server log, so it shouldn't be necessary for developers. It's obviously handy during development, so hence a config option.

Personally, I don't want the stack trace to be displayed for security reasons. Revealing code details and file locations to users may make it easier to attack the code (even if security through obscurity is no security at all)

almarklein:
Ok, I see your point. Thanks.

Can you please update the description of the new config option (now 'Show stack traces in browser window'), to indicate that its about server-side stack traces being shown in the browser?

jrversteegh:
Not really sure about the travis error. TEST_DOCS. looks unrelated.

almarklein:
Yeah that CI build is rather iffy lately :/  this looks good to go. Thanks!

PR #402, opened 2017-09-27
Add flipper example to demonstrate property flickering

almarklein:
This demonstrates a phenomenon that is plaguing us a lot, and one of the things that the proposal in #367 would fix. @korijn @ivoflipse

### DIFF
--- a/flexx/ui/examples/flipper.py
+++ b/flexx/ui/examples/flipper.py
@@ -1,0 +1,167 @@
+"""
+This app demonstrates a phenomenon in the current implementation of Flexx
+that we've referred to as "property flickering" (flipperen in Dutch). This
+occurs when a synced property is repeatedly set from Python. The effect gets
+stronger as Python and/or JS are busier.
+
+In the examples below, pushing the button will add an element to a list
+property. When "one-way mode" is turned on, Python will instead send
+an event to JS, and the adding to the list happens there, causing the
+traffic to be one-directional. This emulates the behavior after the Flexx
+refactoring (actions can be invoked from Python, but mutations only happen in
+JS).
+
+"""
+
+from time import time, sleep
+
+from flexx import app, ui, event
+
+
+
+class FlipperTester(ui.Widget):
+    """ Widget to test the two flipper tests in a single app, and allow switching
+    one-way mode.
+    """
+    
+    def init(self):
+        
+        with ui.VBox():
+            with ui.HBox():
+                ui.Widget(flex=1)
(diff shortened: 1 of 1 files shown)


almarklein:
One can copy-paste this code and run it; I tried to make it somewhat self-explanatory.

ivoflipse:
You seem to be missing an `app.start()` at the end

ivoflipse:
I just played around with it and it becomes very apparent once the Py busyness exceed 0.16 s. As long as we're below that threshold, I can barely click `Add item` fast enough to cause it to slow down, but once you exceed the threshold, things rapidly go out of sync. 

Triggering 8 fast clicks in rapid succession also manages to create a sufficient amount of busyness that it becomes noticeable. Turning on the one-way mode however doesn't seem to solve or avoid this problem.

The batch one-way mode seems to finish processing all the events faster, but it also doesn't feel as responsive, for a few frames nothing happens and then things happen rapidly. In our application, I think users would think the application didn't register their clicks and simply click more often, which is not what we want either

But thanks for the clear example!

Korijn:
Thanks for this proof of concept. I think it's clear how this approach avoids the problematic situation.

Any thoughts on the consequences for objects that cannot easily be sent to JS?

almarklein:
@ivoflipse I think this needs a bit more explanation. The amount of flicker effect depends a lot on the timings. In real apps, like ours, the app will be busy because a change in "items" would spawn many events. I simplified this using a sleep, which may indeed cause large delays in this example, even with one-way mode.

The difference that you should observe in this example, is that without one-way, the number of items (seen from Python) sometimes shrinks before growing again. This causes a lot of extra updates. When this happens on a property that has a lot of downstream events (like in our case), this means the app becoming more busy, and thus again increasing the chance of flicker.

almarklein:
> Any thoughts on the consequences for objects that cannot easily be sent to JS?

In the current situation you can also not set a synced property with data that cannot be send to JS (well, you could, but it would cause an error when Flexx tries to sync to JS).

As for serializing special objects (like binary data), which we now send separately, I'm working on that.

ivoflipse:
Yeah I can confirm that the flickering no longer happens with the one-way communication. 

The example just highlighted the problem I've noticed that once you've build up sufficient state, that the flood of events start to overload the system and performance starts to degrade. I don't have a clear idea how we could prevent that problem right now

almarklein:
Closing. We've seen the effect disappear in real use-cases.

PR #403, opened 2017-10-06
Improve performance of event reconnecting

almarklein:
In one particular use-case, this significantly improves performance.
@ivoflipse @korijn

### DIFF
--- a/flexx/event/_handler.py
+++ b/flexx/event/_handler.py
@@ -399,7 +399,7 @@ class Handler:
             i1 += 1
         # Skip common objects from the end
         i2, i3 = len(new_objects) - 1, len(old_objects) - 1
-        while (i2 > i1 and i3 > i1 and
+        while (i2 >= i1 and i3 >= i1 and
                new_objects[i2][0] is old_objects[i3][0] and
                new_objects[i2][1] == old_objects[i3][1]):
             i2 -= 1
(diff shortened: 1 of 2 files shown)


flexx/event/_handler.py
```diff
@@ -375,22 +375,40 @@ def _connect_to_event(self, index):
...
+        # Prepare disconnecting
+        old_objects = connection.objects  # (ob, type) tuples
+        connection.objects = []
+        
         # Obtain root object and setup connections
         ob = self._ob1()
         if ob is not None:
             self._seek_event_object(index, connection.parts, ob)
-
+        new_objects = connection.objects
```
Korijn: if `connection.objects` returns a reference, your code might not do what you think it does

almarklein: That's why there is a `connection.objects = []` a few lines above :)  (it's a simple attribute, not a property)

Korijn: okay.

flexx/event/tests/test_handlers.py
```diff
@@ -581,4 +581,106 @@ def bar(self, *events):
...
+        f.bars = [Bar(), Bar()]
+    #
+    assert len(triggers) == 0
+    assert len(disconnects) == 0
+    
+    # Change values of bar.spam
+    with event.loop:
+        f.bars[0].spam = 7
+        f.bars[1].spam = 42
+    #
```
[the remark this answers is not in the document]
almarklein: Like whitespace, but still keeping the block together. Sorry for the weird style :)

PR #439, opened 2018-02-13
Updates

almarklein:


flexx/ui/_widget.py
```diff
@@ -346,7 +346,10 @@ def __render_resolve(self, vnode, node=None):
         # Check vnode (we check vnode.children further down)
         if vnode and vnode.nodeName:  # is DOM node
             return vnode
-        if not isinstance(vnode, dict):
+        elif isinstance(vnode, str):
+            vnode = {'type': 'span', 'props': {}, 'children': vnode}
+            # return window.document.createTextNode(vnode)  #  not in node.children
```
Korijn: doesnt createTextNode work?

almarklein: No. I think because it won't appear in node.children, causing the virtual node updating to fail in odd ways.

PR #441, opened 2018-02-13
Changes to not open things up for xss attacks

### LINKED ISSUE #293: Text property (of Button, LineEdit, etc.) is interpreted as HTML
As a consequence, setting the text to e.g. `"<empty>"` or `repr(ob)` will give odd results. It can also be an advantage though, e.g. to include a small image or use markup like `<b>`.

Affected widgets: `Button`, `RadioButton`, `CheckButton`, `LineEdit`, `Label.text` (has probably many cases where HTML is deliberately used). `Widget.title` in some places, `TreeItem.text` and `TreeItem.title`, `ComboBox.items`, `ui.html.XX.text` is deliberately html.

Questions:

* Should we sanitize the input of this prop?
* If not, should we sanitize some of them? E.g. `.title`
* How hard is it to do this in JS? -> easy, look at `html.escape()`, though not sufficient in light of [this](http://wonko.com/post/html-escaping)
* If we sanitize, should we also have a `html` prop?

cc @Maxyme

almarklein:
Closes #293 

This makes it harder to insert raw html in your app. There are still two places where this is possible: the `Label.html` property, where we put a warning in the docs, and via `_render_dom` by providing an `innerHTML` prop, in which case the user probably knows what she is doing, and we warn for it in the docs.

We make use of  `textContent` instead of  `innerHTML` in several places. The `Label` class now has a text and html property, setting either updates both (let the browser do the work, because one can set `textContent` and get `innerHTML` and vice versa).

Also snuck in a tweak to increase the chance of closing the webruntime when Python is killed.

### DIFF
--- a/flexx/app/_asset.py
+++ b/flexx/app/_asset.py
@@ -306,5 +306,5 @@ class Bundle(Asset):
         if len(self.assets + self.modules) > 1:
             source.insert(0, '/* Bundle contents:\n' + '\n'.join(toc) + '\n*/\n')
         #if isjs:
-        #    source.append('window.flexx.spin("%s");' % ('*' * len(self.modules)))
+        #    source.append('window.flexx.spin(%i);' % len(self.modules))
         return '\n\n'.join(source)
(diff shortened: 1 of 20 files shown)


Korijn:
Nice improvements! This makes usage of test vs html much more consistent. :)

PR #466, opened 2018-04-18
Play better with mobile divices, e.g. touch events

### LINKED ISSUE #149: Make "mouse" events work on mobile devices
Many interactions do not work on a touchscreen. This might mean we'd rename some of our standard widget events. So important to get this in before we enter "beta".

almarklein:
Closes #149


ivoflipse:
Nice to have a static example that requires no server!

PR #491, opened 2018-09-20
Improve docs on Widget.keypres event

almarklein:
It seems that the gist of it to always use keydown, unless I miss something.

### DIFF
--- a/flexx/ui/_widget.py
+++ b/flexx/ui/_widget.py
@@ -1058,9 +1058,12 @@ class Widget(app.JsComponent):
 
     @event.emitter
     def key_press(self, e):
-        """ Event emitted when a key is pressed down. This event does not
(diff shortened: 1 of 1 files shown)


Korijn:
keypress is useful when you only want to receive events for keys that actually input a character. it avoids the need to filter oit events for e.g. shift or delete key events.

that's all I can think of.

almarklein:
Yeah, that was what I thought. Unfortunately, Firefox passes arrow keys/backspace/more? and Chrome does not. Well, that's the web :D

Korijn:
Indeed it is.

almarklein:
I guess Firefox' reasoning is to only suppress modifier keys, because these are typically held down, which would cause them to fire in fast succession.

PR #497, opened 2018-10-02
 Add Widget.minsize_from_children property 

almarklein:
Use-case one: in a splitter widget, sometimes you just want to be able to collapse the content all the way.

Use-case two: if a widget inside a splitter widget has a lot of content, you sometimes don't want it to take the full space (and more) of the splitter. 

This adds a `minsize_from_children` property to the `Widget` class that can be set to False, so that only the intrinsic `minsize` of the widget is used to calculate its size constraints.

### DIFF
--- a/flexx/ui/_widget.py
+++ b/flexx/ui/_widget.py
@@ -670,6 +670,8 @@ class Widget(app.JsComponent):
         """
         # Get new limits
         w1, w2, h1, h2 = self._query_min_max_size()
+        w1 = max(0, w1)
+        h1 = max(0, h1)
         # Update the property, so that our parent may react
         self._set_size_limits((w1, w2, h1, h2))
         # Update the style, so that flexbox works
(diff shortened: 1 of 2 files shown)


almarklein:
Example code below. Uncomment that bit to make use of the new feature:

```py
from flexx import flx

class Example(flx.Widget):
    
    def init(self):
        
        with flx.HSplit():
            with flx.VSplit():
                
                with flx.VBox():
                    flx.Button(text='foo')
                    flx.Button(text='bar')
                
                with flx.VBox():
                    flx.Button(text='foo')
                    flx.Button(text='bar')
            
            with flx.VSplit():
                
                with flx.VBox(flex=1): #, minsize_from_children=False, minsize=100, style='overflow-y:auto'):
                    for i in range(40):
                        flx.Button(text=f"button {i}")
                flx.Widget(flex=1)  # spacer

m = flx.launch(Example, 'firefox')
flx.run()
```

PR #504, opened 2018-10-10
openlayers example

ocobacho:
Openlayers example similar to leaflet example but loading some geojson and also allowing to draw points.


flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,177 @@
...
+BASE_DIR = os.getcwd()
+
+with open(BASE_DIR + '/static/js/ol-v4.6.4/ol.css') as f:
+    ol_css = f.read()
+
+with open(BASE_DIR + '/static/js/ol-v4.6.4/ol.js') as f:
+    ol_js = f.read()
+
+flx.assets.associate_asset(__name__, 'ol.css', ol_css)
+flx.assets.associate_asset(__name__, 'ol.js', ol_js)
```
almarklein: Can this not be loaded from a CDN instead?

flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,177 @@
...
...,"features":[{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.08160400390625,23.0548095703125]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-81.31256103515625,22.510986328125]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-80.47210693359375,22.6483154296875]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-80.16448974609375,22.3187255859375]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-81.00494384765625,22.2857666015625]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-81.54876708984375,22.3406982421875]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.56500244140625,23.0767822265625]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.40570068359375,22.9449462890625]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.33428955078125,23.1756591796875]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.30682373046875,23.082275390625]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.21343994140625,22.9669189453125]}},{"type":"Feature","geometry":{"type":"Point","coordinates":[-82.10357666015625,23.192138671875]}}]}'
```
almarklein: This line is too long. This either needs to be written as a multiline string, or perhaps add a ` # noqa` to suppress this style error.

flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,177 @@
...
+        self.map.addLayer(self.vectorLayer)
+
+    @flx.action
+    def add_osm_layers(self):
+        self.map.addLayer(self.baseLayer)
+
+    def _create_dom(self):
+        """ This function gets automatically called when needed; Flexx is aware
+        of what properties are used here.
+        """
```
almarklein: This comment if false. The `_render_dom` is automatically called when any of the properties used within change. The `_create_dom` method is only called once (during initialization).

flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,177 @@
...
+        global document
+        node = document.createElement('div')
+        self.mapnode = document.createElement('div')
+        node.appendChild(self.mapnode)
+        self.mapnode.id = 'maproot'
+        return node
+
+    def _render_dom(self):
+        self.map_init()
+        return super()._render_dom()
```
almarklein: I wonder if you want to call `map_init()` here. The name suggests that you want to call it just once :)

almarklein:
Thanks for this example! I added a few small comments.

flexxamples/howtos/openlayers.py
ocobacho: Yes it can. But with this example I also wanted to show how you can load local assets so your app can work offline and also get feedback if this is the right way to do that? But if you prefer to use a CDN I'll change it no problem.

flexxamples/howtos/openlayers.py
ocobacho: ups, copy paste error. my bad

flexxamples/howtos/openlayers.py
ocobacho: Yes,  You think I should do it on _create_dom?

ocobacho: Thanks for all the useful comments.

flexxamples/howtos/openlayers.py
almarklein: In general I prefer using CDN, because it keeps the Flexx repo small and clean. But I agree that showing the "local variant" also has value, so let's leave it like this :)

almarklein: Could you add a docsting to this example that briefly explains what this howto demonstrates, including how it loads local assets?

flexxamples/howtos/openlayers.py
almarklein: > You think I should do it on _create_dom?

Yes, probably.

almarklein: I think this is the last remaining issue right now, right?

ocobacho: yes, I couldn't make it work inside _create_dom. I putted an if inside map_init so it wont be called every time. Is there something like an after render event?
```suggestion
        if not self.initialised:
```

almarklein: Maybe put it in `init()`? That's usually a good point to initialize stuff, since the DOM and properties have been initialized.

almarklein: Maybe put it in `init()`? That's usually a good point to initialize stuff, since the DOM and properties have been initialized.

almarklein: Maybe put it in `init()`? That's usually a good point to initialize stuff, since the DOM and properties have been initialized.

ocobacho: I've try on the init() without luck. I need to call map_init() after the div that contains the map is loaded on the page. Normally this is done by putting your code after the div. Any Ideas?

ocobacho: I also was thinking about what you said about kipping the repo small  by using CDN. I could use CDN for this example and create another small example demostrating how to load local assets but with smaller css, js files. what do you think?

almarklein: Mmm odd that it does not work in `init()`. Ah well, I guess we can leave it as-is then :)

> I could use CDN for this example and create another small example demostrating how to load local assets but with smaller css, js files.

That would be even better, yes!

ocobacho: ok, I've created a new example and update the last one with CDN.

flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,185 @@
...
+
+from flexx import flx
+import os
+
+BASE_DIR = os.getcwd()
+
+with open(BASE_DIR + '/static/js/data.json') as f:
+    geojson = f.read()
+
+flx.assets.associate_asset(__name__, 'https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.css')
```
almarklein: The line is too long. I thing this should do:
```suggestion
flx.assets.associate_asset(__name__,
    'https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.css')
```

flexxamples/howtos/openlayers.py
```diff
@@ -0,0 +1,185 @@
...
+flx.assets.associate_asset(__name__, 'https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.js')
```
almarklein: Dito:
```suggestion
flx.assets.associate_asset(__name__,
    'https://cdnjs.cloudflare.com/ajax/libs/openlayers/4.6.5/ol.js')
```

almarklein:
Ok, I've merged it. I'll be fixing the CI errors (and the small style error introduced in this PR) in #511.

Thank you for this sweet contribution!

almarklein:
Oh, ignore that comment about the style error. I had overlooked that you already fixed that! My bad!

Issue #525, opened 2018-11-19
Cookies.py example: browser opens after shutdown


maphew:
https://flexx.readthedocs.io/en/stable/examples/cookies_src.html - Running this example in Win10 with python 3.71 seems a little out of step: the browser window doesn't open until after Flexx exits. 

`(oh)` is a conda environment with nothing installed except Flexx and dependencies.

```
(oh) C:\Users\mattw\code\oh
> python cookies.py
[I 16:22:43 flexx.app] Asset store collected 2 new modules.
[I 16:22:43 flexx.app] Serving apps at http://localhost:49191/
[I 16:22:43 flexx.app] Asset store collected 23 new modules.
[I 16:22:44 flexx.app] Starting Flexx event loop.
# nothing happening anywhere at this point
# wait couple minutes, then manually launch Firefox and point at above url
[W 16:24:33 flexx.app] Cleared 1 old pending sessions
[I 16:24:33 flexx.app] New session Cookies ESLJzoqOLzIFgp3gh1l075Ph
# close FF tab, now this happens:
[I 16:25:57 flexx.app] Session closed Cookies ESLJzoqOLzIFgp3gh1l075Ph
[I 16:25:57 flexx.app] Stopping Flexx event loop.
                              
(oh) C:\Users\mattw\code\oh   
>
# **NOW**  Google Chrome launches, with a tab pointed at 
# http://localhost:49191/Cookies/?session_id=fzVkPxmq6zETIyCThx0NU4uC
# but it errors out because Flexx has shut down 
```
Curious things:
- the observed effects aren't exactly the same each time. 1 out of the 6 tests ran while writing up this issue _did_ launch Chrome right away, while Flexx was still running
- Firefox is my default browser, not Chrome
- refreshing the page in either browser causes Flexx to shutdown

almarklein:
> refreshing the page in either browser causes Flexx to shutdown

Right, `flx.start()` should be used instead of `flx.run()` here. Fixed that.

> Firefox is my default browser, not Chrome

That's annoying, but I can indeed reproduce it :(  Tracking this issue at https://github.com/flexxui/webruntime/issues/11

> the observed effects aren't exactly the same each time. 1 out of the 6 tests ran while writing up this issue did launch Chrome right away, while Flexx was still running

That is weird ...  is this only the case for this example, or also for others?

maphew:
[...]

PR #528, opened 2018-11-22
add psutil requirement

maphew:
btw, it's no longer necessary to convert markdown to rst for pypi :-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -111,7 +111,7 @@ setup(
     platforms='any',
     provides=[name],
     python_requires='>=3.5',
-    install_requires=['tornado', 'pscript>=0.6.3', 'webruntime>=0.5.5', 'dialite>=0.5.2'],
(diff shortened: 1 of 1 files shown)


Korijn:
If this is only required for the CPU monitoring demo, I would advise against adding it to the package requirements. Is it used for other purposes?

almarklein:
I agree with @Korijn here. The only use-case is in that (and maybe another) demo.

almarklein:
That said, I appreciate the contribution and don't want to discourage you, @maphew !

almarklein:
Ok, I just read #525 and see the confusion. The correct way to fix this would be to wrap the `import psutil` in a `try-except` and reraise the error with a clear message that one should install `psutil`.

maphew:
No discouragement felt! I was dithering about it being a good idea. In the end I came down on the side of "examples in the docs should just work", but I've no firm opinion about it.

PR #540, opened 2018-12-10
Adds test (and comment) on 'Serving apps at' ready signal

aars:
Here's a start. I guess checking for a url could be done, but that would have to be part of the "spec" then I guess.


almarklein:
I'm trying to think about how this could be improved further, but I guess what's important is that there is *some* message that can be relied on. So let it be "Serving apps at" for the rest of Flexx' lifetime :)

PR #545, opened 2019-03-08
disables useless listeners

aars:


aars:
This shouldn't be a PR. Pressed the wrong button. Sorry.

PR #587, opened 2019-09-01
Fix ipython import for notebooks

jrversteegh:
Maybe you would have preferred an issue report, but since I had to fix it, I thought I'd share the fix instead. The IPython imports fail for me. 
I have to admit I'm not entirely sure I understand the reason for the #noqa style import. Also I don't understand how this has ever worked, so forgive me if I'm just being an idiot here...  
Doing a 'regular' import certainly fixes the issue, but since there must be some reason why you did it this way, I decided to fix it without removing  the `exec`.
See e.g. https://bugs.python.org/issue4831

### DIFF
--- a/flexx/app/_funcs.py
+++ b/flexx/app/_funcs.py
@@ -87,9 +87,9 @@ class NoteBookHelper:
         self.enable()
 
     def enable(self):
-        get_ipython = None
-        exec("from IPython import get_ipython")  # noqa - dont trigger e.g. PyInstaller
-        ip = get_ipython()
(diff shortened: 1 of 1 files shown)


almarklein:
> I decided to fix it without removing the exec.

Thanks. The `exec` was the easiest way to avoid PyInstaller from including IPython (and all its dependencies). It could well be that I did euhm not proprtly test IPython support after this change. Woops.

almarklein:
This can't be the cause for the CI error on Pypy - that must be a glitch.

Issue #593, opened 2019-10-11
GroupWidget rendering issue


matkuki:
Hi, finally found the problem.
This code works without problems (just renders one `Groupwidget` with one `ComboBox`):
```python
from flexx import flx, ui, event, config

class GBX(flx.GroupWidget):
    CSS = f"""
        .flx-GBX legend {{
            font-family: Helvetica;
            font-size: 16px;
            font-weight: bold;
        }}
    """

class Test(flx.Widget):
    def init(self):
        with flx.VBox(padding=0, spacing=10):
            with flx.HBox(padding=0, spacing=0):
                    with GBX(title="SOME LONG GROUPBOX TITLE", flex=1):
                        with flx.VBox():
                            flx.ComboBox(
                                editable=False,
                                options=["ONE", "TWO", "THREE"],
                                selected_index=0,
                            )
            flx.Widget(flex=1)
test_app = flx.App(Test, title="TEST")
test_app.serve('')
flx.start()
```
![image](https://user-images.githubusercontent.com/10289651/66683449-bd239580-ec77-11e9-9140-f41519e914c6.png)

But change the line `class Test(flx.Widget):` to `class Test(flx.PyWidget):`, and about 4 out of 10 times the `GroupWidget` layout is messed up like so:
![image](https://user-images.githubusercontent.com/10289651/66683488-db899100-ec77-11e9-9fd6-9b4ad749a72b.png)

P.S.:
Simplified the example code.

jrversteegh:
I can reproduce this. Could you try to make your example minimal? I noticed you can remove a lot without losing this behaviour..

jrversteegh:
It seems like the order of the inline css is different for PyWidget than for Widget.

For the simplest example the order with Widget is:
 * _widget.css
 * _layout.css
 * _hv.css
 * _dropdown.css
 * _group.css

While for PyWidget:
 * _widget.css
 * _group.css
 * _layout.css
 * _hv.css
 * _dropdown.css

Manually adding a copy of the contents of _group.css at the end fixes the issue. Even if you only add the `.flx-GroupWidget > .flx-Layout` rule. Does this rule only work properly after styling of .flx-Layout? This is beyond my CSS knowledge. It seems like a good idea though that css is loaded in the same order regardless of Widget or PyWidget to avoid these kind of issues @almarklein

almarklein:
[...]

PR #595, opened 2019-10-14
Mark GroupWidget determined dimensions !important

jrversteegh:
* GroupWidget determined dimensions should take precedence. #593


almarklein:
Yes, I think `!important`, though generally to be avoided, is the way to go here.

almarklein:
Not sure why the pypy build is suddenly failing, but it can hardly be this change, so I'm merging.

almarklein:
Thanks @jrversteegh !

PR #596, opened 2019-10-14
Fix unintended combobox clearing

### LINKED ISSUE #591: ComboBox selection disappears
Hi,

When creating `ComboBox` like so:
```python
flx.ComboBox(
    editable=True,
    options=my_list_of_options,
    selected_index=0,
)
```
and then:
- selecting an option
- clicking anywhere inside the `ComboBox` (with or without doing any editing)
- clicking outside the `ComboBox`

the text inside the `ComboBox` disappears!
Here is a video:

This seems like a bug. Is this supposed to behave like this?

jrversteegh:
Closes #591


almarklein:
Awesome! LGTM

PR #604, opened 2019-11-06
Update _filebrowser.py

matkuki:
If there are no subdirectories in the current viewed path, clicking on the top file does nothing, as the emitted directory is not `None` but is invalid (it's an invalid value: **path + "/" + name_of_last_directory**).
This is a quick fix for this issue.

I do not have enough experience to know how to change that the first file emits `"directory": None` (like the other files do), but if it's simple then tell me and I'll update the code.

### DIFF
--- a/flexx/ui/pywidgets/_filebrowser.py
+++ b/flexx/ui/pywidgets/_filebrowser.py
@@ -165,7 +165,7 @@ class FileBrowserWidget(PyWidget):
         dirname = events[-1].dirname
         filename = events[-1].filename
         print(dirname, filename)
-        if dirname:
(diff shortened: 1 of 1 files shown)


almarklein:
Ah, I think that its because at line 81 and 84, we only set `dirname` or `filename`. We should set both (the one that we don't set now should be set to `None`).

matkuki:
Updated it as you described and also tested it and it works.
Can you check it?

almarklein:
Looking good, thanks!

almarklein:
Pypy is broken on Travis, will have a look at that in a bit.

PR #607, opened 2019-11-15
Add GitHub URL for PyPi and use README.md as long_description

andriyor:
Warehouse now uses the `project_urls` provided to display links in the sidebar, as well as including them in API responses to help automation tool find the source code for flexx. For example, see Django's [setup.cfg](https://github.com/django/django/blob/master/setup.cfg) and [PyPI listing](https://pypi.org/project/Django/).

PyPI also already support rendering project descriptions from Markdown https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi/

Also if you don't want to execute` __init__.py` for receiving `__version__` you can use setup.cfg for that purpose as did it at Django https://github.com/django/django/pull/12013

### DIFF
--- a/flexx/__init__.py
+++ b/flexx/__init__.py
@@ -1,32 +1,3 @@
-"""
-`Flexx <https://flexx.readthedocs.io>`_ is a pure Python toolkit for
-creating graphical user interfaces (GUI's), that uses web technology
-for its rendering. Apps are written purely in Python; The
-`PScript <https://pscript.readthedocs.io>`_ transpiler generates the
-necessary JavaScript on the fly.
-
-You can use Flexx to create (cross platform) desktop applications, web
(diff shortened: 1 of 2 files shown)


almarklein:
Awesome, thanks!

One remark: the idea is that the docs in the `__init__.py` are a bit of a summary. They are re-used both for package `long_description` and as an intro on the actual docs (the ones on RTD). The readme has somewhat more info that's common on a Github README.

I prefer to keep the info on Pypi brief, i.e. use the docs from the `__init__.py`. Or perhaps alternatively, we can add a delimiter in the readme (using an HTML comment, perhaps) and the code in `/flexx/docs/conf.py` can take its summary from there.

almarklein:
Closing for inactivity. I've added `project_urls` in the mean time though.

PR #628, opened 2020-03-09
Allow colon passed to apply_style()

Abeldewit:
Because of the CSS parsing, apply_style() would split on any colon which would split http:// links. With this small change it will only split on the first colon and then on the next semicolon for a next line of css


almarklein:
Thanks!

PR #659, opened 2020-07-30
Typo

Konubinix:
The intent was to set the selected_key to '', but the code
mistakenly use the selected_key as a function.


Konubinix:
I configured emacs to remove the trailing whitespaces on save, this is why there are several diffs about removing some extra spaces.

I did not remove then from the patch, for I guessed we would agree it would be a good thing to remove them anyway. But if you like, I can provide a patch with only the functional diff.

almarklein:
Thanks! Nice catch! Yeah, removing the whitespace is fine.

PR #664, opened 2020-08-19
Issue when validating the existance of an asset when the asset is a link

ceprio:
I add an issue when using `flx.assets.associate_asset(__name__, 'https://cdn.plot.ly/plotly-latest.min.js')` in my __main__ module. Turns out the asset store was not detecting the name in it's list of loaded assests as the name in the dict is shorten to the filename.

I applied correction an now submitting. 

Thanks,


almarklein:
Thanks, this looks to be a suitable fix!

PR #685, opened 2021-01-14
fix deprecated lib call

btwe:
encodestring is deprecated since 3.1,
was still available until 3.8,
gone in 3.9
encodebytes since 3.1, too


almarklein:
Thanks! I should add 3.9 to CI as well.

PR #698, opened 2021-05-26
Symmetry betwee PyWidget and Widget

ceprio:
Hi almarklein, I would like to push a change to have more symmetry between PyWidget and Widget:
I can do:
```
    with ui.Widget(style=overlay_style+"width:0%;") as self.waiting_overlay:
        (...)
```
But could not do:
```
    with ui.PyWidget(style=overlay_style+"width:0%;") as self.waiting_overlay:
        (...)
```
Which I found anoying as I'm using server side processing quite a lot and need to switch Widget from one side to the other seamlessly. 

I changed the `bokehdemo.py` and inserted a PyWidget within to show the style in action.

I have a lot of other things relying on this change so I hope it goes forward one way or another. I hope to submit them at a later point when they get more reliable.

P.S. there is also an optimisation for flask serving within the code (in _flaskserver.py).

Thank you for your time.

### DIFF
--- a/flexx/event/_component.py
+++ b/flexx/event/_component.py
@@ -206,6 +206,8 @@ class Component(with_metaclass(ComponentMeta, object)):
         # With self as the active component (and thus mutable), init the
         # values of all properties, and apply user-defined initialization
         with self:
+            # This call will pop some of the consumed property values
+            # will remain properties used for the DOM component initialisation
             self._comp_init_property_values(property_values)
             self.init(*init_args)
 
@@ -218,9 +220,9 @@ class Component(with_metaclass(ComponentMeta, object)):
     def _comp_init_property_values(self, property_values):
         """ Initialize property values, combining given kwargs (in order)
         and default values.
+        Property values are popped when consumed so that the remainer is used for 
+        other initialisations without mixup.
         """
-        # Property values must be poped when consumed so that the remainer is used for 
-        # instantiation of the Widget
         values = []
         # First collect default property values (they come first)
         for name in self.__properties__:  # is sorted by name
--- a/flexx/event/_component.py
+++ b/flexx/event/_component.py
@@ -203,7 +203,7 @@ class Component(with_metaclass(ComponentMeta, object)):
         for name in self.__properties__:
             self.__handlers.setdefault(name, [])
 
-        # With self as the active component (and thus mutatable), init the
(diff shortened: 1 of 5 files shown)


flexx/ui/_widget.py
```diff
@@ -1158,7 +1158,9 @@ def _comp_init_property_values(self, property_values):
         # when this is the active component, and after the original
         # version of this has been called, everything related to session
         # etc. will work fine.
-
+        # Property values must be poped when consumed so that the remainer is used for 
+        # instantiation of the Widget
```
almarklein, line 1162: Could you move this comment down to where `super()._comp_init_property_values(property_values)` is called to say that this call will pop the values ?

flexx/event/_component.py
```diff
@@ -219,6 +219,8 @@ def _comp_init_property_values(self, property_values):
         """ Initialize property values, combining given kwargs (in order)
         and default values.
         """
+        # Property values must be poped when consumed so that the remainer is used for 
+        # instantiation of the Widget
```
almarklein, line 223: Could you move this comment to the docstring above for extra visibility? And we can just say that they are popped so that the remainder can be used for something else. We're in a low-level part of Flexx here, and this code knows nothing of widgets :)

almarklein:
Wow, thanks for this! You're brave to wander in the most complex parts of Flexx ;)

I took a good look and verified that (as far as I can see) popping these values have no consequences elsewhere. I have just a few comments. These are more nitpicky than I'd normally do, but since this code is complex it's worth being extra careful about comments.

flexx/event/_component.py
ceprio: Done

flexx/ui/_widget.py
ceprio: I've added the comment all along the trail... It should be clear now. :)

ceprio:
Hi almarklein, the last commit integrates the requested changes on the comments. 

Flexx is complex but that is what you get when trying to separate GUI from code. Is there a better way? I do think your solution is elegant and brings portability for the created widgets. That is the main driving force behind Flexx.

Thank you,

almarklein:
I meant that these low-level component classes are among the most complex code, but this solution is done very well :)

Anyway, thanks!

PR #700, opened 2021-06-07
include 39 in ci

almarklein:


almarklein:
*sigh* Travis connection is gone. I'd really like to move to GH actions, but don't have the time right now.

PR #705, opened 2021-09-02
Fix CI glitches

almarklein:


almarklein:
That seems to do the trick.

PR #718, opened 2022-01-13
Added a DynamicWidgetContainer

ceprio:
Hi Almarklein, 

I've added a dynamic container for widgets. I'm using it for a few applications and I thought of sharing it as I see some questions regarding adding and removing widgets on the fly.

Look at the examples, it will show you what it is capable of.

Side note: The file flexx/event/_js.py was not working for me, there is some error in _clean_code_of_thread_stuff. It seem to block a lot of examples so you are probably on it. Just use the old version or the fixed one if you have one.

Have fun ;)


flexx/ui/pywidgets/_dynamicwidgetcontainer.py
```diff
@@ -0,0 +1,81 @@
...
+            self.pages[events[0]['page_position']] = None
+
+    @event.emitter
+    def remove(self, page_position):
+        return dict(page_position=page_position)
+
+    @event.reaction("_emit_instantiate")
+    def __instantiate(self, *events):
+        with self:
+            with events[0]['widget_type'](events[0]['style']) as page:
```
almarklein, line 42: Could you please add a few lines of comments explaining the flow. As I understand it, first a `threading.task` is appended to `.pages` and that is exchanged for an actual page instance here. Does this trick have to do with thread-safety?

flexxamples/howtos/dynamic_container_in_background.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @flx.reaction("but.pointer_click")
+    def delete_function(self, *events):
+        self.parent.remove(self.dyn_id)
+        self.parent.instantiate(PyWidget1)
+
+
+class Example(flx.PyWidget):
+
+    # The CSS is not used by flex in PyWiget but it should be applied to the top div: TODO
```
almarklein, line 60: There are a few TODO's in this this PR's code. Do you plan to fix these in this PR, or do you think that's not needed?

flexxamples/howtos/dynamic_container_app.py
```diff
@@ -0,0 +1,80 @@
+from flexx import flx
```
almarklein, line 1: Could you please add a short docstring at the top of the examples to briefly explain the purpose of the examples?

almarklein (changes requested):
Thanks! This looks like a nice contribution!

If I understand correctly, this new class is needed, because although the JS Widget supports dynamically changing parents, this does not really work for the PyWidget, right? (Sorry, I have not actively used Flexx for a while).

I left a few comments, but none are big issues, I think.

flexxamples/howtos/dynamic_container_app.py
ceprio: Yes, will do.

ceprio:
> If I understand correctly, this new class is needed, because although the JS Widget supports dynamically 
> changing parents, this does not really work for the PyWidget, right?

Yes, that is right. My goal is to have a flexx server that can dynamically serve a client application. The GUI of the client can change completely according to the user actions: view it as a web page that a user is browsing where the pages are dynamically generated by the flexx application.

I'm currently working with that class and will try to put a better sample at one point. Especially the sample with the thread which is intended for a single client application and that should be extended to serve multiple browser clients.

I will work on the requested changes (Though not right away for some of them; let me play more with the base class and come back with better samples).

